PR #1774, opened 2026-08-11
fix: warn honestly before ending a Windows security or servicing process

### LINKED ISSUE #1773: [Bug]: Process Manager — The kill warning promises "will not crash Windows" for Defender and Windows Installer
## Problem

v1.59.0 correctly stopped treating provenance as criticality: `IsKernelCritical` had refused **any** process whose `SafetyLevel` was `System`, which is 59 of the 108 database entries — including `notepad`, `calc` and `explorer`. The app told users that ending Notepad "would cause a system crash (BSOD)". Dropping that arm was right.

But the single warning that replaced the refusal is not true for all 49 processes it now covers. Computed against `Data/ProcessDescriptions.json` (`safety == "System"` minus the 13 boot-critical names), the newly-killable set includes:

```
MsMpEng, NisSrv, SecurityHealthService, SecurityHealthSystray   <- Defender / AV
TrustedInstaller, msiexec, WmiPrvSE, WUDFHost                   <- servicing / installers
Registry, System, Idle, audiodg, spoolsv, sihost, explorer, ...
```

All of them get this text (`ProcessManagerViewModel.cs:220-223`):

> Ending it will not crash Windows, but a feature may stop working or look broken until you sign out or restart.

For `explorer` that is accurate. For the other two groups it is a false reassurance from the app itself:

- **Security** — ending `MsMpEng` is an antivirus-disable step. It will fail on a protected-process OS, but the app should not be reassuring about the attempt.
- **Servicing** — `ProcessManagerService.KillProcess` uses `Kill(entireProcessTree: true)` (`ProcessManagerService.cs:122`). Killing `TrustedInstaller` or `msiexec` mid-operation can leave a half-applied update or a corrupt component store. That damage is *not* undone by the restart the message offers as the remedy.

The confirmation gate itself is correct — this is a wording/tiering defect, not a missing guard.

## Steps to reproduce

1. Open Process Manager.
2. Find `MsMpEng.exe` (Antimalware Service Executable) or `TrustedInstaller.exe`.
3. Click End task.
4. The prompt says ending it "will not crash Windows" and that at worst "a feature may stop working ... until you sign out or restart".

## Expected behavior

A third tier between refusal and the ordinary Windows-component warning: name the real risk (protection switched off, an update interrupted part-way, damage that a restart does not repair) and still let the user decide. Boot-critical stays refused; ordinary Windows components keep the current message; third-party apps keep the plain confirm.

## Affected tab

Process Manager.

laurentiu021:
Follow-up to my own change in v1.59.0. That fix was right — the app had been refusing to end Notepad and telling the user it would blue-screen the machine, because `IsKernelCritical` treated the database's `SafetyLevel == "System"` (provenance) as criticality, and 59 of the 108 entries carry it.

Dropping that arm made **49** entries killable. Computed fresh from `Data/ProcessDescriptions.json` (`safety == "System"` minus the 13 boot-critical names), that set includes:

```
MsMpEng, NisSrv, SecurityHealthService, SecurityHealthSystray   <- security
TrustedInstaller, msiexec, WmiPrvSE, WUDFHost                   <- servicing
Registry, System, Idle, audiodg, spoolsv, sihost, explorer, ... <- the other 41
```

Every one of them got the same message:

> Ending it will not crash Windows, but a feature may stop working or look broken until you sign out or restart.

For `explorer` that is accurate — the taskbar comes back on its own. For the other two groups it is the app reassuring the user about something it should not:

- **Security.** Ending `MsMpEng` is a step towards switching off the machine's protection. It fails on a protected-process OS, but the prompt should not be calm about the attempt.
- **Servicing.** `ProcessManagerService.KillProcess` uses `Kill(entireProcessTree: true)` (`ProcessManagerService.cs:122`). Ending `TrustedInstaller` or `msiexec` mid-write can leave a half-applied update or a corrupt component store — damage that **survives the restart the message offers as the remedy**.

The confirmation gate was never missing. This is a wording and tiering defect: one sentence could not be true for 49 different processes.

## The three tiers

| Tier | Behaviour |
|---|---|
| `BootCriticalProcesses` (13 names) | **Refused outright** — unchanged |
| `HighConsequenceProcesses` (8 names, new) | Confirmed, with a prompt naming the real damage and stating a restart does not undo it |
| Any other Windows component | The existing "a feature may look broken until you sign out" warning — unchanged |
| Everything else | The plain "unsaved work may be lost" confirm — unchanged |

Still a confirmation, deliberately **not** a second refusal list. These really can be ended, and it is the user's machine — what changes is that the prompt tells the truth about the cost. Nothing that could be ended before has become harder to end.

Both sets match on process **name**, not on the database's `safety` field, for exactly the reason `IsKernelCritical` does: that field is provenance, and a process absent from the database (or renamed in it) must not silently fall out of a tier.

## Guards

- Six `[Theory]` cases assert the security/servicing prompt appears **and** that the `"will not crash"` reassurance does not — the negative half matters, since that string is what made the old message wrong.
- One asserts the tier still **asks** rather than refusing, so it cannot quietly drift into a second denylist.
- `HighConsequenceAndBootCritical_DoNotOverlap` — the tiers are evaluated in order, so a name in both sets would be refused and its warning never shown: config that looks present while being dead. Both sets are hand-maintained string lists, so this is asserted rather than assumed.

## Verification

**Red control** (`origin/main`, detached worktree):
```
  [FAIL] MsMpEng.exe gets the honest warning — still promises "will not crash Windows ..."
  [FAIL] NisSrv.exe ... [FAIL] SecurityHealthService.exe ...
  [FAIL] TrustedInstaller.exe ... [FAIL] msiexec.exe ... [FAIL] WmiPrvSE.exe ...
  [PASS] explorer.exe keeps the ordinary Windows-component message
  [PASS] a third-party app keeps the plain confirm
  [PASS] lsass.exe is refused before any prompt
=== 6 FAILED ===
```

**Green** (this branch), run twice: **ALL PASS (10/10)** — including the three controls above, which prove the new branch did not leak into the other tiers.

Safety of the harness itself: every prompt is **declined**, so `KillProcess` is never reached, and the PID used (`999999`) does not exist. No process on this machine was touched.

- Leak scan: **0 hits** across all 32 terms.
- No debug code, no generic catches added.

## Docs

- README's Process Manager section described two tiers; it now describes all three.
- ARCHITECTURE.md records the tier structure **and why both sets match by name rather than by the `safety` field** — that reasoning is the thing a future change is most likely to get wrong, since undoing it is what caused the original bug.

Closes #1773

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -74,7 +74,13 @@ QA-verified is marked with `IsInDevelopment` (surfaced as a PREVIEW badge) inste
 - `StartupViewModel` — startup program management (enable/disable via registry).
 - `DuplicateFileViewModel` — duplicate file finder with partial-hash pre-filter.
 - `DiskAnalyzerViewModel` — disk space breakdown by folder with drill-down.
-- `ProcessManagerViewModel` — running processes with kill, filter, sort.
+- `ProcessManagerViewModel` — running processes with kill, filter, sort. The kill path has three
+  tiers, because one message cannot be true for all of them: `BootCriticalProcesses` (13 names) is
+  refused outright, `HighConsequenceProcesses` (Defender's engine, Windows Installer and the servicing
+  processes) is confirmed with a prompt naming the real damage, and any other Windows component gets
+  the "a feature may look broken until you sign out" warning. Both sets are matched by process NAME,
+  not by the description database's `safety` field — that field is provenance, and treating it as
+  criticality is what made the app refuse to end Notepad while claiming a BSOD.
 - `BatteryHealthViewModel` — charge %, health %, wear, cycle count via WMI.
 - `UninstallerViewModel` — winget-based app uninstaller with batch support.
 - `PerformanceViewModel` — per-tweak performance tuning with snapshot restore.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.61.2] - 2026-08-11
+
+Ending a Windows security or update process now warns you about what actually happens,
+instead of saying a feature might look broken until you sign out.
+
+### Fixed
+- **The "End task" warning was too reassuring for some Windows processes.** Version 1.59.0 fixed the opposite problem — the app used to refuse to end Notepad and claimed doing so would blue-screen your PC — but the message that replaced the refusal said the same thing for every Windows process: that it "will not crash Windows" and at worst "a feature may stop working until you sign out or restart". That is true for something like Explorer, which comes back on its own. It is not true for Windows Defender's engine, where closing it is a step towards switching off your protection, and it is not true for Windows Installer and the servicing processes, where interrupting one part-way through can leave an update half-applied — and restarting does not put that right. Those now get their own prompt that says so plainly. The processes Windows genuinely cannot survive losing are still refused outright, ordinary Windows components keep the existing warning, and normal programs keep the plain one, so nothing you could end before has become harder to end.
+
 ## [1.61.1] - 2026-08-11
 
 Two more of the project's own tests can no longer touch your real settings: the Settings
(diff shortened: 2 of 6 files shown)


laurentiu021:
Superseded by #1782. This branch accumulated three rounds of CHANGELOG and csproj conflicts as other fixes merged ahead of it, and one of those resolutions — performed by GitHub's own "update branch" merge — silently dropped the v1.61.2 entry for a version that had already shipped. Rather than keep reconciling, the fix was replanted clean onto current main: identical code, re-verified end to end (red control 6 failures, green 10/10, all four projects 0 errors / 0 warnings), and its changelog entry joins the unreleased 1.61.3 section instead of claiming a new version number — which the version gate added in #1781 now requires.

PR #1825, opened 2026-08-12
fix: refuse an update target that is not this application's own executable

### LINKED ISSUE #1823: [Bug]: Updater — the self-update applier trusts its target path from the command line
## Problem

Two paths in the self-update mechanism reach the update **applier** while dropping the verification the normal install path performs. Both were verified against current `main` (`d52b73b`).

### 1. `--apply-update` validates the target path's shape, not its identity

`App.xaml.cs:65` takes the applier branch **before** the single-instance mutex, before DI and before the CLI guard:

```csharp
if (UpdateApplier.TryParseArgs(e.Args, out var targetExe, out var oldPid))
{
    LogService.Init();
    UpdateApplier.Run(targetExe, oldPid);
```

`UpdateApplier.TryParseArgs` (`UpdateApplier.cs:71-81`) checks only: the sentinel matches, `args[1]` is non-blank, `args[2]` parses as a positive int. `targetExe` is never compared against the running executable, its directory, or any trusted root. `Run` then calls `ApplyCopy(Environment.ProcessPath, targetExe)`, whose final step is `File.Move(staging, targetExe, overwrite: true)` (`:122`), and afterwards `Process.Start(targetExe)` with `UseShellExecute = true`.

So a command line of the form `SysManager.exe --apply-update "<any writable path>" <pid>` makes the app copy its own 85 MB executable over that path and then launch it. `WaitForProcessExit` returns immediately for a non-existent pid (`ArgumentException` is caught at `:222`), so no wait is needed to reach the write.

`BuildArguments` (`:58-64`) rejects a `"` in the path, but that constrains only the string this app *emits*; the inbound branch re-establishes none of it, because an attacker supplies the command line directly rather than calling `BuildArguments`.

Whatever token launched the process bounds the damage — but the app's documented workflow is "Run as administrator", and `Run`'s own comment (`:203`) notes that `UseShellExecute` deliberately inherits the applier's elevation.

### 2. Rollback executes a user-writable binary with no verification

`AboutViewModel.RollBackAsync` (`:787-840`) launches `%LOCALAPPDATA%\SysManager\updates\SysManager-previous.exe` after a single `File.Exists` check (`:790`), with `UseShellExecute = true` (`:819-824`).

The install path immediately above it (`InstallUpdateAsync`, `:678-728`) is hardened exactly against this shape: it opens a `FileShare.Read` handle **before** verifying, checks SHA-256 from that held stream, then checks Authenticode, and holds the handle across `Process.Start` — its own comment at `:678-683` describes closing the TOCTOU window on "a user-writable `%LOCALAPPDATA%` file".

Rollback shipped later (#1765) and reuses `BuildArguments` plus the applier — documented at `:779-782` as reusing "the SAME applier path" — but inherits none of that verification. `PreserveCurrentBuild` wrote the file, so its provenance is trusted; provenance is not integrity when the directory is user-writable and the file persists indefinitely between the write and the click. No hash is stored beside it to check against.

## Why the existing tests did not catch either

- `UpdateApplierTests.TryParseArgs_RejectsMalformedInput` asserts only that *malformed* args are rejected. No test asserts that a **well-formed but foreign** `targetExe` is refused.
- `AboutViewModelTests` rollback coverage (`:527-593`) asserts presence/absence and label text. No test asserts a tampered `SysManager-previous.exe` is refused.

This is the "validation is the sole defense" category — except for #1 the validation does not exist yet.

## Proposed fix

1. **Applier:** require `targetExe` to be a path SysManager could legitimately occupy — reject anything that is not an existing file whose name is the app's own executable name, and refuse a target under a system root. Refuse loudly (log + exit non-zero) rather than silently.
2. **Rollback:** write a hash beside the retained build in `PreserveCurrentBuild` and verify it before launch, using the same held-handle pattern as `InstallUpdateAsync` so verify-then-execute is not two independent opens.
3. A regression test per path, each asserting the hostile input is **refused**, per the "if a validation is the sole defense it must be explicitly tested" rule.

## Affected tab

About (Updates), and app startup.

laurentiu021:
Closes #1823

## Problem

Two paths reach the update **applier** while dropping the verification the normal install path performs. Both verified against `main`.

### 1. `--apply-update` validated the target's shape, not its identity

`App.xaml.cs:65` takes the applier branch **before** the single-instance mutex, DI and the CLI guard, so `targetExe` arrives straight from `argv`. `TryParseArgs` checked only: sentinel matches, `args[1]` non-blank, `args[2]` a positive int. Nothing compared the target against the running executable. `Run` then calls `ApplyCopy(Environment.ProcessPath, targetExe)`, ending in `File.Move(staging, targetExe, overwrite: true)`, and launches the result with `UseShellExecute = true` — which its own comment notes deliberately inherits elevation.

**Reproduced end-to-end against unfixed code:** a temp file containing `127.0.0.1 localhost` came back **162,304 bytes** after `UpdateApplier.Run(victim, 999999)`. `WaitForProcessExit` catches `ArgumentException` for a non-existent pid and returns immediately, so no wait is needed to reach the write.

`BuildArguments` rejecting `"` constrains only the string this app *emits*; an attacker supplies the command line directly, so the inbound side must re-establish trust itself.

### 2. Rollback launched a user-writable binary on `File.Exists` alone

`RollBackAsync` started `%LOCALAPPDATA%\SysManager\updates\SysManager-previous.exe` after one existence check, with `UseShellExecute = true`. That directory is user-writable and the file persists indefinitely between being written and the click, so anything running as the user could replace it and have it start with SysManager's token.

`InstallUpdateAsync` directly above hardens this exact shape — `FileShare.Read` handle opened **before** verifying, SHA-256 from the held stream, then launch without reopening. Rollback (#1765) reused `BuildArguments` and the applier, documented as "the SAME applier path", but inherited none of the verification.

## Fix

**`IsValidApplyTarget`** — the gate lives in `Run`, the only method that writes, so a future second caller cannot bypass it. Requires:
- a canonicalised path, so validation and the write act on the same string (`sub\..\hosts` is judged as `hosts`);
- this executable's own name, derived from `Environment.ProcessPath` rather than hardcoded;
- a location outside the Windows / System / Program Files roots, compared on a directory boundary;
- an **existing** file — an update replaces an install, it never creates one, which is what stops the applier planting a new binary at a chosen path.

It returns a *reason*, so a refusal is logged specifically instead of looking identical to a successful update.

**Rollback** — `PreserveCurrentBuild` records the SHA-256 of the bytes it wrote. `TryOpenVerifiedPreviousBuild` opens with deny-write **before** hashing and **returns the handle**, so the caller launches the bytes it verified rather than reopening the path. Fails closed: a missing or unusable checksum is a refusal, matching what `UpdateService` does for a missing published `.sha256`. `CanRollBack` now requires the checksum too, so a build retained by an earlier version is not offered rather than offered-then-refused.

## Regression proof (red → green)

`dotnet test` is not run on this workstation, so a harness drove the **real methods** via `InternalsVisibleTo` — behaviour, not source shape.

```
===== RED =====
  [FAIL] the victim file was NOT overwritten by the application binary — 162304 bytes
  [FAIL] UpdateApplier.IsValidApplyTarget exists
  [FAIL] UpdateApplier.TryOpenVerifiedPreviousBuild exists
  ... (11 total)
=== 11 FAILED ===

===== GREEN =====
  [PASS] the victim file was NOT overwritten by the application binary — 19 bytes
  [PASS] an existing copy of this executable is accepted
  [PASS] a path inside the Windows directory is refused FOR ITS LOCATION — the target is inside a protected system directory
  [PASS] a real existing system binary is refused on its name — the target is not named …
  [PASS] a retained build swapped after saving is REFUSED — the saved version has changed since it was saved
  [PASS] a retained build with no checksum is refused — there is no recorded checksum for the saved version
=== ALL PASS ===
```

19 bytes = the original `hosts` content, untouched.

15 unit tests in `UpdateApplierTests`, each asserting the hostile input is **refused**, with positive controls so "refuse everything" cannot score full marks:

- `ApplyTarget_AcceptsAnExistingCopyOfThisExecutable` (the control)
- `ApplyTarget_RefusesAFileThatIsNotThisExecutable` — `hosts`, `evil.exe`, `SysManager.dll`, `notepad.exe`
- `ApplyTarget_RefusesAPathThatDoesNotExistYet`
- `ApplyTarget_RefusesAPathInsideTheWindowsDirectory` — asserts specifically the *location* reason
- `ApplyTarget_RefusesARealExistingSystemBinary` — a file that genuinely exists, so it cannot pass via the existence check
- `ApplyTarget_ResolvesDotSegmentsBeforeDeciding`
- `Run_WithAHostileTarget_LeavesThatFileUntouched` — end-to-end, hash-compared, and asserts not even a `.new` staging file appeared
- `PreserveCurrentBuild_RecordsTheChecksumOfWhatItWrote`
- `VerifiedPreviousBuild_AcceptsAnUntouchedRetainedBuild` — also proves the returned handle denies writes
- `VerifiedPreviousBuild_RefusesABuildSwappedAfterItWasSaved`
- `VerifiedPreviousBuild_FailsClosedWithoutAUsableChecksum` — absent / empty / garbage / wrong-length
- `VerifiedPreviousBuild_RefusesWhenThereIsNoRetainedBuild`

The name and system-root branches are each pinned as **independently reachable**: the location check runs before the existence check, so it cannot be shadowed, and a test proves each fires for its own reason.

## Verification

- All four projects build Release: **0 warnings, 0 errors** (rebuilt after rebasing onto 1.64.1).
- Leak scan over the full diff against `leak-terms.txt`: 0 hits.
- Version 1.64.2 across the triple; CHANGELOG entry with the plain-English lead in this same PR.
- No behaviour change to a normal update: it always pointed at the running executable, which is exactly what the new rule accepts.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,17 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.64.2] - 2026-08-12
+
+Two ways the built-in updater could be tricked into damaging your PC are now closed off. Nothing about
+normal updating changes.
+
+### Security
+- **SysManager could be told to overwrite the wrong file.** When SysManager installs an update, it starts the newly downloaded copy with a hidden instruction naming the file to replace. That instruction was only checked for the right shape, not for what it actually pointed at — so anyone who could start SysManager with a crafted command line (a shortcut, a scheduled task, a script someone talked you into running) could name any file on your PC and have SysManager write over it. If you were running SysManager as administrator at the time, that meant almost any file, including ones Windows needs. SysManager now refuses unless the file it is told to replace really is an existing copy of SysManager itself, outside protected system folders, and writes a note in its log saying why it refused. Normal updates are unaffected; they always pointed at the right file.
+- **"Go back to the previous version" now checks the saved copy before starting it.** SysManager keeps one copy of your previous version so you can go back after a bad update, and that copy sits in your own user folder where other programs can also write. Starting it was previously based only on the file being there — so anything running on your PC could have swapped that copy for something else and had SysManager launch it, with administrator rights if SysManager was elevated. SysManager now records a checksum when it saves the copy and verifies it right before starting it, keeping the file locked in between so it cannot be swapped in the meantime. If the copy has changed, or was saved by a version before this one and has no checksum, the button is not offered and you are told why instead of it happening silently. Installing an update already worked this way; going back now does too.
+
+---
+
 ## [1.64.1] - 2026-08-12
 
 A rare crash that could happen if you closed the window while the Resource History tab was loading
--- a/SysManager/SysManager.Tests/AboutViewModelTests.cs
+++ b/SysManager/SysManager.Tests/AboutViewModelTests.cs
@@ -523,6 +523,19 @@ public sealed class AboutViewModelRollbackTests : IDisposable
 
     private string PreviousBuild => Path.Combine(_dir, UpdateApplier.PreviousBuildFileName);
 
+    /// <summary>
+    /// Writes a retained build the way the applier does — the binary AND its checksum. A rollback is
+    /// only offered when both exist, because SysManager will not start a saved build it cannot verify,
+    /// so writing the binary alone sets up the legacy/tampered state rather than the healthy one.
+    /// </summary>
+    private void RetainBuild(string content = "OLD-BUILD")
+    {
+        File.WriteAllText(PreviousBuild, content);
+        File.WriteAllText(
+            UpdateApplier.PreviousBuildHashPath(_dir),
+            UpdateApplier.ComputeFileHash(PreviousBuild));
+    }
+
     [Fact]
     public void CanRollBack_IsFalse_WhenNoPreviousBuildWasRetained()
     {
@@ -538,11 +551,25 @@ public sealed class AboutViewModelRollbackTests : IDisposable
     [Fact]
     public void CanRollBack_IsTrue_WhenAPreviousBuildExists()
     {
-        File.WriteAllText(PreviousBuild, "OLD-BUILD");
+        RetainBuild();
 
         using var vm = NewVm();
 
         Assert.True(vm.CanRollBack);
+    }
+
+    [Fact]
+    public void CanRollBack_IsFalse_WhenTheRetainedBuildHasNoChecksum()
+    {
+        // A build retained by a version that predates the checksum, or one an attacker dropped in.
+        // SysManager will not start a saved build it cannot verify, so the offer must not appear at all
+        // — being offered a button that then refuses is worse than not seeing it.
+        File.WriteAllText(PreviousBuild, "OLD-BUILD");   // binary only, deliberately no checksum
+        Assert.False(File.Exists(UpdateApplier.PreviousBuildHashPath(_dir)));
+
+        using var vm = NewVm();
+
+        Assert.False(vm.CanRollBack);
     }
 
     [Fact]
@@ -558,7 +585,7 @@ public sealed class AboutViewModelRollbackTests : IDisposable
         // The file can disappear between the button appearing and being pressed (manual cleanup, disk
         // tools, another instance). That must produce an explanation and a corrected UI rather than a
         // silent no-op or a crash.
-        File.WriteAllText(PreviousBuild, "OLD-BUILD");
+        RetainBuild();
         using var vm = NewVm();
         Assert.True(vm.CanRollBack);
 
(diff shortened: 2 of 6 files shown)


laurentiu021:
CI was red for a real reason — two of my own regressions, both correct failures:

- `CanRollBack_IsTrue_WhenAPreviousBuildExists` and `RollBack_WhenThePreviousBuildVanished_SaysSo_AndHidesTheOffer` hand-wrote only `SysManager-previous.exe`. Now that `CanRollBack` requires the recorded checksum too, those fixtures were setting up the *legacy/tampered* state rather than the healthy one. They now write both via a `RetainBuild` helper mirroring the applier, and a new test pins the state they used to occupy by accident: **a retained build with no checksum must not be offered at all.**

Two improvements that came out of it:

- **The verification now runs before the confirmation dialog.** Asking the user to approve a rollback and then refusing it is the wrong order. It also means the deny-write handle is already held while the prompt is on screen, so the file cannot be swapped during those seconds. The `finally` still releases it on the cancel path.
- **Added `RollBack_WhenTheRetainedBuildWasSwapped_RefusesAndExplains`** — binary present, checksum present, bytes changed. That is what a same-user attacker replacing the saved copy looks like, and it asserts the refusal reaches the UI instead of the launch proceeding.

One thing worth flagging separately, found while diagnosing this and **not caused by this PR**: the run also logged 5 `dotnet format` errors in `DuplicateFileViewModelTests.cs`, `DarkModeViewModelTests.cs` and `AboutViewModel.cs:114/119` — files this PR does not touch. They are present on `main` and #1822 logged the identical five yet passed, because the "Check formatting" step runs four `dotnet format` commands in one `run:` block, so only the **last** project's exit code decides the step. The style drift and the gate that lets it through are both real; I am handling them separately rather than bundling unrelated formatting into a security fix.

Harness re-verified green (23/23) after the restructure; both projects build 0 warnings / 0 errors.

PR #1826, opened 2026-08-12
docs: correct the supported-version table and gate it in CI

laurentiu021:
## Problem

Three documentation claims were wrong in ways a reader would act on. All re-derived from source, not taken from a report.

### 1. `SECURITY.md` named a supported version three minors stale (highest impact)

```
| 1.61.x   | :white_check_mark: |
| < 1.61   | :x:                |
```

Shipping version is 1.64.x. The policy says "security fixes are applied to the latest minor release only" and then names 1.61.x as that release — so a user on the current build matches **neither row** and reasonably reads their own version as unsupported. This is on the page that decides whether someone bothers to report a vulnerability at all.

### 2. `ARCHITECTURE.md:36` named a helper method that does not exist

It said `BuildNavGroups()` uses `Group()` and `Item()`. Grep for a nav-item helper in `MainWindowViewModel.cs` returns `EagerItem` (line 166) and `Group` (line 269); `Tab<TVm>()` is the third. There is no `Item()`. Someone adding a tab greps for `Item(`, finds nothing, and cannot tell whether to use the lazy `Tab<TVm>()` or `EagerItem()` — a distinction the same document explains at length under Dependency Injection.

### 3. `ARCHITECTURE.md:139-144` sent readers to a file that does not contain what it claims

The seam list said each entry is "registered against its implementation and mock-substitutable (see `ServiceRegistration.cs`)". Derived from that file, **12** interfaces are registered. Two named or expected seams are not:

- **`IDialogService`** is in the list but registered nowhere — it is consumed through the static `DialogService.Instance`, which is exactly why its swapping tests must live in the serialized xUnit collection (a fitness function enforces that).
- **`IBandwidthMonitorService`** — the one seam with two shipping implementations (`ConnectionBandwidthSource`, `EtwBandwidthSource`), described in detail later in the same document — was absent from the list, and is injected as factories.

## Fix

- Corrected the table to 1.64.x, and added a sentence pointing at the releases page: the newest minor is what's supported, and being on a build newer than the table never means unsupported. That removes the "stay quiet" reading even during the window before a docs edit lands.
- **Gated it mechanically.** The existing "Check version consistency" step now also asserts SECURITY.md's supported minor matches the version being released. A prose number that nothing recomputes will drift again; this makes it impossible to release past it. Only the MINOR line is checked, so frequent patch releases don't each demand a docs edit.
- Rewrote both ARCHITECTURE.md passages to describe what the code actually does, including *how* the two non-registered seams are reached, so the next reader isn't sent to the wrong file.

## Verification

- **The gate was proven both ways before committing** — not just added:
  - green against the corrected table (`derived minor: 1.64` / `SECURITY says: 1.64`);
  - **red** against a copy still saying 1.61.x (`RED: gate correctly rejects the drift`).
  A gate that cannot go red is theatre.
- Both workflow files parse as YAML after the edit.
- Every replacement claim re-derived from current source: 12 registered interfaces from `ServiceRegistration.cs`, the two item helpers from `MainWindowViewModel.cs`, the shipping version from the csproj.
- Leak scan over the diff against `leak-terms.txt`: 0 hits.

`docs:` — no release triggered.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,12 +58,32 @@ jobs:
       # Format-check every C# project the workflow builds, not just the main one —
       # the same .editorconfig governs the test projects, so otherwise their style
       # could drift past CI silently.
+      #
+      # `shell: bash` with an explicit failure sweep, NOT four bare lines. A multi-line `run:` uses the
+      # default shell, where only the LAST command's exit code decides the step — so a project that
+      # failed formatting printed its errors as annotations and the step still went green. That is how
+      # five real violations (DuplicateFileViewModelTests, DarkModeViewModelTests, AboutViewModel)
+      # reached main across several PRs while every run looked clean: the UITests project happened to
+      # pass last. Each project's status is now captured and the step fails if ANY of them did.
       - name: Check formatting
+        shell: bash
         run: |
-          dotnet format SysManager/SysManager/SysManager.csproj --verify-no-changes --verbosity diagnostic
-          dotnet format SysManager/SysManager.Tests/SysManager.Tests.csproj --verify-no-changes --verbosity diagnostic
-          dotnet format SysManager/SysManager.IntegrationTests/SysManager.IntegrationTests.csproj --verify-no-changes --verbosity diagnostic
-          dotnet format SysManager/SysManager.UITests/SysManager.UITests.csproj --verify-no-changes --verbosity diagnostic
+          failed=""
+          for proj in \
+            SysManager/SysManager/SysManager.csproj \
+            SysManager/SysManager.Tests/SysManager.Tests.csproj \
+            SysManager/SysManager.IntegrationTests/SysManager.IntegrationTests.csproj \
+            SysManager/SysManager.UITests/SysManager.UITests.csproj
+          do
+            if ! dotnet format "$proj" --verify-no-changes --verbosity diagnostic; then
+              failed="$failed $proj"
+            fi
+          done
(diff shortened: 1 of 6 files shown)


laurentiu021:
Added a second, related fix found while diagnosing a red run on #1825.

**The formatting gate could not fail on three of its four projects.** "Check formatting" ran four `dotnet format` commands as four lines of one multi-line `run:` block; in the default shell only the **last** command's exit code decides the step. So a project that failed printed its violations as `##[error]` annotations and the step still went green.

Not theoretical — **five real violations are on `main` right now** because of it:

- `DarkModeViewModelTests.cs` — import order (`NSubstitute` before `System.IO`)
- `DuplicateFileViewModelTests.cs:191,195` — object-initializer wrapping
- `AboutViewModel.cs:114,119` — constructor body placement

`SysManager.UITests` runs last and passes, so several merged PRs logged the identical five errors and were reported green. I noticed because #1825 surfaced them and I initially took them for my own.

The step now captures each project's status and fails if any failed, naming them. **Verified the logic goes red when the first project fails and the last passes** — precisely the shape that used to pass.

The five violations are fixed in the same commit by `dotnet format` itself, because the corrected gate would otherwise turn this PR red for pre-existing drift. Formatting only, no semantic change; all four projects verified format-clean and building 0 warnings / 0 errors afterwards.

This belongs here rather than in a separate PR because this PR already changes the same `ci.yml` gate block, and splitting them would leave `main` with a gate that reports errors it cannot act on.

PR #1905, opened 2026-08-17
fix: read Volume Control peak meters off the UI thread in one batched call

### LINKED ISSUE #1831: [Bug]: Volume Control — peak meters do audio-engine calls on the UI thread 20 times a second
## Problem

The Volume Control tab's peak meters run on a `DispatcherTimer` at 50 ms (`AudioMixerViewModel.cs:29,60-61`), so `UpdatePeaks` executes **on the UI thread** 20 times a second while the tab is visible.

`UpdatePeaks` (`:179-183`) loops every row and calls `_service.GetPeak(row.SessionId)`. `AudioMixerService.GetPeak` (`:276-308`) takes `lock (_gate)` and then, per cached audio interface in the group, makes an `IAudioMeterInformation.GetPeakValue` cross-apartment COM call into the Windows audio engine.

So per second, on the render thread: rows × streams marshalled COM calls, plus one `PropertyChanged` per row driving a `ProgressBar` invalidation (`AudioMixerView.xaml:104`).

It also contends for `_gate` with the 1 Hz reconcile loop, which holds that same lock while enumerating every render session (`EnumerateGroupsLocked`, `:109`). So roughly once a second a peak tick blocks on the UI thread behind a full session enumeration.

## Why it matters

With ten apps playing audio this is ~200+ COM transitions per second on the thread that draws the window, and one of every twenty ticks waits on the reconcile lock. The symptom is choppy meters and sliders that feel laggy on a modest laptop — the same class of defect as the Bandwidth Monitor fix in 1.61.9, but on a 20× hotter cadence.

The service itself already treats this path as hot: the `COMException` handler at `:290-299` documents that one failing session used to write ~20 identical log lines per second, calling it "the log's single biggest flood path".

## What is already correct (so the fix does not undo it)

The timer **is** visibility-gated — `OnIsActiveChanged` starts it at `:285` and stops it at `:288`, so nothing polls while the tab is hidden. The reconcile loop is correctly off-thread (`Task.Run(_service.GetSessions)`, `:119`) and the slow identity resolution is deliberately kept outside the COM lock. The defect is only the on-UI-thread cost while the tab is open.

## Steps to reproduce

1. Play audio in several apps.
2. Open the Volume Control tab and drag a slider while the meters animate.

## Proposed fix

Sample peaks off the UI thread and marshal only the resulting values back, the way the Bandwidth Monitor was fixed — with the post-await disposal re-check that fix needed (see 1.63.1, and 1.64.1 for the same class in Resource History).

Worth considering alongside it: 50 ms is faster than a meter needs to be legible; a slower interval would cut the cost proportionally with no perceptible loss. That is a judgement call for the tab's owner, not something to change silently.

## Affected tab

Volume Control.

laurentiu021:
Closes #1831

## The defect

The Volume Control VU meters refreshed 20 times a second from a `DispatcherTimer` whose `Tick` called `IAudioMixerService.GetPeak` **once per visible row, synchronously, on the render thread**. Each of those calls took the service lock and marshalled a cross-apartment COM `GetPeakValue` per audio control in the session group.

With ten apps playing that is **~200 COM transitions a second on the thread that draws the window** — and roughly one tick in twenty blocked there behind the 1 Hz reconcile, which holds the same lock while enumerating every session. Symptom: meters that jittered instead of moving smoothly, and sliders that lagged behind the mouse.

Same defect class as the Bandwidth Monitor fix in 1.61.9, at 20x the cadence. Present since the tab shipped in 1.52.37.

## The fix

- **`IAudioMixerService.GetPeaks(IEnumerable<string>)`** — returns a value for every id asked for (0 for one that is gone). `AudioMixerService` materialises the ids **before** taking the lock (the caller's sequence is a LINQ projection over a bound collection, so enumerating it inside the lock would run UI-thread collection reads under the gate the reconcile also wants), then does all the COM work in one acquisition. `GetPeak` and `GetPeaks` share one `PeakLocked` core — a single COM read path, not two that can drift.
- **`PeakLoopAsync` replaces the `DispatcherTimer`**, sharing the reconcile loop's cancellation token so `Dispose` stops both. Each tick awaits one `Task.Run(GetPeaks)` and writes only the finished numbers back on the UI thread.
- **`UpdatePeaksAsync` re-checks `_reconcileCts` and the token after the await.** The sample now genuinely yields, so a `Dispose` landing mid-sample would otherwise re-light meters it had just zeroed — the use-after-dispose shape pinned in 1.63.1 (Bandwidth) and 1.64.1 (Resource History). `Dispose` now cancels *before* zeroing, for the same reason.
- `OnIsActiveChanged` no longer starts/stops a timer; the loop itself skips its work while the tab is hidden.

## Red/green proof (mutation)

Baseline written pristine and verified green first, then three mutations. 41/41 tests green unmutated.

| Mutation | `ReadsEveryRowInOneServiceCall` | `WritesEachRowPeak` | `WhenDisposedMidSample` |
|---|---|---|---|
| *(unmutated)* | green | green | green |
| Revert to a **per-row `GetPeak` loop** (the pre-fix shape) | **RED** | green | green |
| Sample synchronously, no worker thread | green | green | green |
| Drop the post-await disposal guard | green | green | **RED** |

The first row is the point of the batch: the old code produced **correct numbers**, so only a call-count assertion can catch a regression to it. The `Peaks()` test helper therefore stubs **both** peak reads from one level table on purpose — otherwise a per-row shape would read an unstubbed method, return 0 everywhere, and fail the value assertions, making the call-count test look redundant when it is the only thing actually pinning the batching.

Row 3 is recorded for honesty, not as a guard: no unit test here proves the work is off-thread (that needs a Dispatcher). Row 4 confirms the dispose test reddens for its own reason and nothing else.

Both waits in the dispose test are **bounded (10s)**. On the correct shape neither is reached — the stub is entered on the worker thread and released explicitly, so no assertion depends on wall-clock timing. They exist because a wrong shape would deadlock the whole suite instead of failing one test: a shape that never calls `GetPeaks` never sets the signal, and a shape that samples on the *calling* thread parks the test inside the stub before the in-flight task is even assigned. (Both were observed while building this proof.)

## Propagate check

Grepped every remaining `DispatcherTimer` view model and read each tick body: `StandbyMemoryViewModel` (2s, one `GlobalMemoryStatusEx`), `DarkModeViewModel` (60s registry read), `DisplayProfileViewModel` (countdown decrement), `NetworkSharedState` (250ms UI *flush* — marshaling to the UI thread is its purpose). None does per-row cross-apartment COM under a contended lock. Volume Control at 50ms was the last instance of this class.

## Verification

- `dotnet build -c Release` — app + Tests + IntegrationTests + UITests: **0 errors, 0 warnings** (all four)
- Guard harness: **41/41 green**
- Leak scan: all 32 terms from `leak-terms.txt` against the full diff — **0 hits**
- Author headers present on all four touched source files
- `ARCHITECTURE.md` corrected — it claimed a `DispatcherTimer` drives the meters
- CHANGELOG entry in this same PR, with the plain-English lead paragraph
- csproj bumped to 1.65.11


laurentiu021:
## Follow-up commit: the UI suite caught a regression in this fix

CI's UI job reported 2/135 failing. The job is `continue-on-error`, so the check still shows **pass** — the failures are only visible in the log. Investigated both rather than accepting the green tick.

**`UninstallerUiTests.CurrentSession_ShowsMatchingGuidanceWithoutAdminRelaunchButton` — not mine.** It fails identically on three dependabot branches that ran today and touch no app code (`32008138883`, `32008111800`, `32008106726`, each 1/135 with this exact test). Environment-dependent, pre-existing, tracked separately.

**`FunctionalUiTests.Drivers_List_ProducesCountOrDone` — mine.** Only on this branch, and Drivers has no audio dependency — but the mechanism is real:

- The UI suite shares **one app process** across classes (`[Collection("App")]`), and `AllTabsSmokeUiTests` visits `nav-volume-control`. So once that smoke test runs, this view model is alive for the rest of the suite.
- The `DispatcherTimer` I removed was genuinely **stopped** on deactivate. My loop only checked `IsActive` *after* its delay — so it woke 20 times a second for the whole life of the process, and since the continuation resumes on the captured context, each no-op wake queued a work item onto the **Dispatcher**.
- Every later test then ran against that background churn. `Drivers_List_ProducesCountOrDone` has the tightest wall-clock bound in the suite (25 s around an external enumeration), so it was the one that fell over.

That is a genuine regression against the timer's behaviour, not a flaky test. Fixed in `8039967`: `PeakLoopAsync` now awaits an **activation gate** before its delay — a `TaskCompletionSource` completed on show, replaced with a fresh incomplete one on hide — so a hidden tab costs nothing at all. `WaitAsync(ct)` keeps Dispose able to exit the loop while parked.

### Proof for the gate

| Mutation | `HiddenTab_ParksThePeakLoop` | `Deactivating_ClearsPeaks` |
|---|---|---|
| *(unmutated)* | green | green |
| Delete the re-arm on hide (one-shot gate = the tick-and-skip shape) | **RED** | green |
| Loop ignores the gate, bookkeeping kept | green | green |
| Unconditional re-arm (behaviour-preserving variant) | green | green |

Row 3 is recorded for honesty: the test asserts the gate's **state machine**, including the re-show path — a one-shot gate would leave the meters dead after the user navigated away once — but it does not prove the loop consumes it. Row 4 confirms it doesn't over-constrain the implementation.

Re-ran the batching proof against the gated loop: per-row still reddens **only** the call-count test, and the disposal guard still reddens only the dispose test. (One change: the "synchronous sample" mutation now also reddens the dispose test, because the gate makes that test's `IsActive = true` release the loop — strictly more discriminating than before.)

Guard harness 42/42 green; all four projects 0 errors / 0 warnings; leak scan 0 hits on the new delta. CHANGELOG, ARCHITECTURE and the class doc now state the park-while-hidden behaviour instead of implying a skip.

PR #1907, opened 2026-08-17
test: assert the Uninstaller copy the app ships, and guard the whole class

laurentiu021:
## The defect

`UninstallerUiTests.CurrentSession_ShowsMatchingGuidanceWithoutAdminRelaunchButton` failed **1/135 on every CI run of 2026-08-17** — including three dependabot branches that touch no app code. My first read was "environmental". It isn't.

PR #1808 rewrote the Uninstaller's elevated banner:

| | text |
|---|---|
| app ships (since #1808) | "Uninstalling is turned off while SysManager runs as administrator, so each app can show you its own permission prompt…" |
| test asserted | "Uninstall is disabled in administrator sessions. Reopen SysManager normally to continue." |

The second string exists **nowhere** in the app. The assertion could never pass.

Two things hid it for weeks:

1. The elevated branch only runs when the test session is elevated — as on the CI runner, **not** on a developer's box.
2. The `ui-tests` job is `continue-on-error`, so `gh pr checks` prints `UI automation tests  pass` while the log says `Failed: 2`. The job's `conclusion` is `success` too, so `gh run view --json jobs` is equally blind.

## The fix

**The instance:** the test now waits for a stable **fragment** of each banner rather than a whole sentence, so a future reword doesn't break it. Its failure message now reports which integrity-level branch ran and the text it wanted, so the next person doesn't have to reverse-engineer that from a bare "did not show the guidance".

**The class:** `ArchitectureTests.EveryUiTextAssertion_QuotesCopyTheAppActuallyShips` — every multi-word phrase a UI test waits for must exist somewhere in the app's XAML or C#. It lives in the **blocking unit suite** deliberately: the defect it pins is precisely the one the non-blocking UI job cannot report loudly enough to stop a merge, and detecting it needs no desktop session — it is text comparison over two source trees.

Scope came from enumerating the population, not guessing:

- Checked when a literal is an argument of a text-waiting helper, **or** a ternary branch building expected copy. That second shape matters: the original defect assigned via a ternary and passed the *variable*, so a call-site-only regex reported **zero** findings on a tree that genuinely had one.
- Three words or longer. A 20-character floor was tried first and excluded nearly every real call site, because the durable waits (`"CPU"`, `"total"`, `"drivers found"`) are deliberately short.
- Matching the call *argument* rather than the whole line keeps assertion-failure messages out of scope even when they share a line, as in `Assert.True(_fx.HasText("Dashboard"), "Dashboard did not recover…")`.
- Vacuity floor of 7, set after listing the 8 real phrase literals. It earned its place immediately: it caught both earlier versions of this guard, which parsed only 3 and then 8-minus-the-ternary.

## Red/green proof (mutation)

Baseline written pristine and verified green first.

| Mutation | guard |
|---|---|
| *(unmutated)* | green |
| Restore the historical Uninstaller wording (the real defect) | **RED** |
| Reword the app's banner, leave the fixed test alone (the *future* shape) | **RED** |
| Rename the local the guard once keyed on **and** use stale copy | **RED** |
| Rename the local only, wording still correct | green |

Rows 3 and 4 are the pair that matters: the guard is not evadable by renaming the variable it was written around, and it does not false-alarm on a harmless refactor.

## Sweep

All 44 text assertions across the UI suite were checked against the app's rendered strings. This was the **only** genuine drift — the other candidates were assertion-failure messages, correctly excluded.

## Verification

- All four projects `-c Release`: **0 errors, 0 warnings**
- Guard harness over `ArchitectureTests`: my guard green; the 4 remaining reds are harness limitations (three `[Theory]` methods invoked with no args, one helper that walks up from the harness's own `bin`), not code defects — CI runs them correctly
- Leak scan: all 32 terms from `leak-terms.txt` against the diff — **0 hits**
- `test:` — no version bump, no CHANGELOG entry, no release (the pre-commit hook enforces this pairing, and correctly blocked the first attempt when the branch was still named `fix/`)

### DIFF
--- a/SysManager/SysManager.Tests/ArchitectureTests.cs
+++ b/SysManager/SysManager.Tests/ArchitectureTests.cs
@@ -2303,6 +2303,150 @@ public partial class ArchitectureTests
             + $"holds:\n  {string.Join("\n  ", notBound)}");
     }
 
+    /// <summary>
+    /// Every sentence a UI test waits for must be copy the app actually ships. A UI test that quotes
+    /// wording nothing renders can never pass — it is a permanently red assertion masquerading as
+    /// coverage.
+    /// </summary>
+    /// <remarks>
+    /// <para>Found by a sweep after <c>UninstallerUiTests</c> failed 1/135 on every CI run of one day,
+    /// including branches that touch no app code. PR #1808 rewrote the Uninstaller's elevated banner and
+    /// left the test asserting the previous sentence, which then existed nowhere in the app. Two things
+    /// hid it: the branch only runs when the test session is elevated (as on the CI runner, not on a
+    /// developer's box), and the <c>ui-tests</c> job is <c>continue-on-error</c>, so <c>gh pr checks</c>
+    /// printed "pass" while the log said <c>Failed: 2</c>.</para>
+    /// <para>This guard lives in the BLOCKING unit suite on purpose. The defect it pins is one the
+    /// non-blocking UI job cannot report loudly enough to stop a merge, and it needs no desktop session
+    /// to detect — it is pure text comparison over the two source trees.</para>
+    /// <para>Scope is chosen by what can actually rot. A one-word wait like <c>"CPU"</c> or a fragment
+    /// like <c>"drivers found"</c> survives rewording and is not checked; a multi-word PHRASE is what a
+    /// copy edit breaks, so the bar is two spaces (three words) rather than a character count — the
+    /// original defect's threshold experiment showed a 20-character floor excludes nearly every real
+    /// call site, because the durable waits are deliberately short. Literals carrying markup, path, or
+    /// format-hole characters are ids, xpaths and templates, not copy. Matching is whitespace-normalised
+    /// and case-insensitive because XAML wraps attribute values across lines.</para>
+    /// </remarks>
+    [Fact]
+    public void EveryUiTextAssertion_QuotesCopyTheAppActuallyShips()
+    {
+        var uiTestsDir = Path.Combine(FindRepoRoot(), "SysManager", "SysManager.UITests");
+        Assert.True(Directory.Exists(uiTestsDir),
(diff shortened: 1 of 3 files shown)


laurentiu021:
## Correction, and a second fix

The Uninstaller assertion now **passes** — that part is confirmed. But the UI job still reported `Failed: 1`, and it is `Drivers_List_ProducesCountOrDone`, on a branch that touches only test files.

That means my batch-83 attribution was **partly wrong**. I credited the earlier Drivers failure entirely to the Volume Control peak loop (PR #1905). The activation gate there did improve the margin, but it was not the whole root cause.

The evidence, per run:

| run | Uninstaller | Drivers | Drivers duration |
|---|---|---|---|
| `31811949758` (main) | FAIL | pass | — |
| `32008750468` (peak meters, 1st) | FAIL | **FAIL** | 25 s |
| `32012084750` (peak meters, gated) | FAIL | pass | — |
| `32019870080` (this PR) | **pass** | **FAIL** | 27 s |

Both Drivers failures land at or just past the 25 s bound, and it passes on the runs in between — it fails **at the wall**, not on a wrong result.

**Root cause:** the scan spawns `pwsh` and queries `Win32_PnPSignedDriver`, one of WMI's slowest classes, with a cold PowerShell start on a shared runner. A 25 s budget for that is optimistic; the test has been passing by luck.

**Fix (`d69b809`):**
- Budget raised to 90 s — still far inside the UI job's 30-minute ceiling, and the whole suite runs in about 2.5 minutes.
- The failure message now distinguishes **still scanning** from **went quiet**. The progress text is only on screen while the scan is in flight, so its presence at timeout means a slow environment and its absence means the command finished without setting a summary. Previously this reported a bare `Assert.NotNull() Failure: Value is null`, which says nothing about which happened.
- Deliberately **not** accepting a bare `"Done"` as a third terminal signal, even though the view model sets it: `WaitForText` searches the whole window, so it would match another tab's status line and let the test pass without the scan finishing.

No app code changed in either commit — this remains `test:`.

PR #2043, opened 2026-09-01
fix: stop the Dashboard stating disk and memory verdicts it never measured

laurentiu021:
Two Dashboard tiles asserted facts about the user's hardware that the app had not checked. On a maintenance utility read by a non-technical user as a verdict on their own machine, that is the worst kind of defect available.

## P1 — "All SMART indicators healthy" for drives that were never read

`DiskHealthService.Collect` swallows `ManagementException`, `UnauthorizedAccessException` and `COMException` and returns its **partially-filled** list. So a Storage WMI namespace that is broken, disabled, or access-denied reaches `HealthScoreService` as an **empty list**, never as an exception.

`ComputeDiskScore` returned `100` for that, and `ScanSmartHealthAsync` branches on `>= 90`. Result: a machine whose disk health could not be read at all was told its disks were in good health.

The same shape applied to `ComputeRamScore` and `ComputeUptimeScore` (100 for a null snapshot), while `BuildRecommendations` suppressed everything because each arm requires a non-null source. So a machine where all three reads failed showed **a perfect score with no advice**, and nothing anywhere said why.

The file already argued the opposite principle twelve lines below the bug:

> 80, not 100, for a drive we know nothing about: absent evidence is not a clean bill of health, **and this is the only case that reaches it**.

That last clause was false. The `Count == 0` path is a *stronger* absence of evidence and was scored *more generously*.

## P2 — "No memory errors (30 days)" was reading memory usage

The tile classified on `RamScore`, which `HealthScoreService` derives entirely from `snapshot.Memory.UsedPercent` — a usage percentage with no relation to hardware errors.

The service that answers this question already exists, is a registered singleton, and its own doc reads *"Scans the System event log for hardware-error events (WHEA) in the last 30 days"*: `MemoryTestService.CheckErrorLogsAsync`. It was consumed **only** by `SystemHealthViewModel`, which words the same result correctly.

Both directions were wrong:

- real WHEA errors at 40% memory use → green **"No memory errors (30 days)"**;
- a healthy machine at 85% memory use → yellow **"Memory issues detected"**, which the System Health tab then contradicted.

## Change

- An unread component scores `UnknownComponentScore` (80) — the same value the per-drive unknown rule already used — so it can never reach a green branch.
- `HealthScoreResult.UnavailableComponents` lets a caller say "could not be read" instead of reading a verdict out of a fallback number. Unknown is **not** "degrading" either: nothing is degrading, nothing was measured, so it gets its own wording.
- The memory tile calls `CheckErrorLogsAsync` and words its outcomes to match `SystemHealthViewModel`, so the two surfaces cannot disagree about the same 30 days. A failed scan reports as unknown rather than as good news.
- Both alert decisions are now pure internal statics, matching `ClassifyAppUpdates` in the same file. Previously both lived inside a `Dispatcher.BeginInvoke` callback where no test could reach them.
- `MemoryTestService` injected into `DashboardViewModel`; the one production construction site and all four test sites updated.

## Three existing tests required the defect

`ComputeDiskScore_NullDisks_Returns100`, `ComputeDiskScore_EmptyList_Returns100` and `ComputeRamScore_NullSnapshot_Returns100` asserted that a missing source scores perfect. That requirement *was* the bug, written down — the same shape as the dead `SeverityToLevel` twin in the previous batch. All three are rewritten to pin the rule and renamed after it. The uptime arm had **no test at all**, which is presumably how it survived two rewrites of the other two; it has one now.

## Red before, green after

**Baseline: all 16 green** (12 new/rewritten + 4 pre-existing).

| Mutation | Expected red | Actual |
|---|---|---|
| **A** — an unread disk scores 100 again | the two disk tests | matched |
| **B** — the unavailable branch removed | the SMART test | matched |
| **C** — WHEA errors made invisible | 2 memory tests | matched |
| **D** — a failed scan reported green | the scan-failed test | matched |
| **E** — `UnknownComponentScore` raised to 95 | the test that owns that number | matched |

**Mutation E first exposed a flaw in my own tests.** The four arm tests asserted `Equal(UnknownComponentScore, ...)` — comparing the constant to itself — so raising it moved both sides of the assertion and only one test noticed. The number now has a dedicated test pinning the literal `80` *and* its relationship to the Dashboard's `>= 90` green threshold; the arm tests keep asserting "this arm returns the unknown score", which is their actual job.

Mutation C also had to be rewritten: `if (false)` is `CS0162` unreachable code, an error under warnings-as-errors, and a mutation that cannot compile proves nothing. A threshold no real machine reaches has the same effect.

Both touched files were restored byte-for-byte after each mutation and rebuilt from the restored source.

## Regression sweep

- `dotnet build -c Release`: app, unit tests **and** integration tests — all **0 warnings, 0 errors**.
- `dotnet format --verify-no-changes`: all three projects exit 0.
- Leak scan over the whole branch diff against `main`: 43 patterns, 313 added lines, **0 hits**.
- Local CI gates re-run from the extracted workflow bodies: version consistency (1.75.20 across csproj, CHANGELOG, SECURITY.md), valid bump (v1.75.19 → 1.75.20), CHANGELOG plain-English lead. All pass.
- `README.md` and `ARCHITECTURE.md` do not quote either tile's text, so neither needed a change.

## Scope

The audit's fuller suggestion — nullable component scores with the weights redistributed over the components that actually reported, as the battery arm already does — is a better model. It also changes `HealthScoreResult`'s shape, the CLI's JSON contract (`CliRunner.cs:146`) and the gauge's inputs. That is a separate change; this one makes the missing evidence representable and stops the false verdicts, which is the confirmed defect.

CHANGELOG entry and version bump to 1.75.20 are in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,32 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.75.20] - 2026-09-01
+
+Two Dashboard tiles were telling you things SysManager had not actually checked. If your drive health could
+not be read, the tile still said your drives were healthy. And the "No memory errors" tile was not looking at
+memory errors at all — it was looking at how much memory was in use.
+
+### Fixed
+- **"All SMART indicators healthy" no longer appears when your drives could not be read.** Reading drive
+  health needs a part of Windows that is sometimes broken, switched off, or blocked without administrator
+  rights. When that happened SysManager got nothing back — and treated nothing back as a perfect result, so
+  the tile reported your drives were in good health without having looked at them. It now says the health
+  could not be read and points at the System Health tab. A drive that genuinely reports a problem is
+  unaffected: those wordings and thresholds are unchanged.
+- **The "No memory errors (30 days)" tile now actually looks for memory errors.** It was deciding based on how
+  much memory was currently in use, which has nothing to do with whether your memory has faulted. So a
+  computer with real memory hardware errors was told there were none as long as memory use was low, and a
+  perfectly healthy computer running a lot of programs was warned about "memory issues" that the System Health
+  tab then said did not exist. The tile now reads the same 30-day Windows error record the System Health tab
+  reads, and words the result the same way, so the two can no longer disagree. If that check itself cannot
+  run, it says so rather than reporting good news.
+- **The overall health score no longer counts unread information as perfect.** Drive health, memory and uptime
+  each scored full marks when their source returned nothing, so a computer where all three failed showed a
+  perfect score with no advice at all. An unread component is now scored as unknown, which is what a single
+  unreadable drive was already scored — the code said so in a comment while doing the opposite one screen
+  above.
+
 ## [1.75.19] - 2026-09-01
 
 The Logs tab could get stuck loading forever, with one processor core pinned at full speed and the Refresh
--- a/SysManager/SysManager.IntegrationTests/AllViewModelsSweepTests.cs
+++ b/SysManager/SysManager.IntegrationTests/AllViewModelsSweepTests.cs
@@ -32,7 +32,7 @@ public class AllViewModelsSweepTests
     private static CrashMarkerService TempCrashMarkers()
         => new(Path.Combine(Path.GetTempPath(), "SysManagerTests", "sweep-crash"));
 
-    [Fact] public void Dashboard_Constructs() => Assert.NotNull(new DashboardViewModel(new SystemInfoService(), new TuneUpService(new ShortcutCleanerService(), new DiskHealthService(), new SystemInfoService()), new HealthScoreService(new SystemInfoService(), new DiskHealthService(), new BatteryService()), new TemperatureService(new DiskHealthService(), skipHardwareInit: true), new WingetService(new PowerShellRunner()), TempCrashMarkers()));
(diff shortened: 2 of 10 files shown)


laurentiu021:
CI caught a real error of mine, worth recording rather than quietly force-pushing over.

**`ComputeUptimeScore_NullSnapshot_Returns100` failed** — 1 failed, 4952 passed. The PR body said *"The uptime arm had **no test at all**"*. That was wrong: the test exists at `HealthScoreServiceTests.cs:127` and pinned the defect exactly like its three siblings. My claim came from a grep I truncated with `head -12`, so the fourth match never reached the screen, and I added a duplicate test under a new name instead of fixing the one that was there.

Fixed: the pre-existing test is rewritten with its three siblings, and my duplicate is removed — two tests asserting the same thing under different names is how a suite starts overstating its own coverage.

So it is **four** existing tests that required the defect, one per arm, not three plus a gap.

Also swept for anything else in the same shape. The remaining `Returns100` tests are `ComputeDiskScore_AllHealthy`, `ComputeRamScore_LowUsage`, `ComputeRamScore_60Percent`, `ComputeUptimeScore_1Day` and the three `ComputeBatteryScore` cases. All of those assert a **measured** perfect result, which is correct and untouched — and the battery arm never had this bug, because it redistributes its weight when there is no battery instead of scoring one.

Verified locally after the fix: the 8 perfect-score and unknown-score tests all green, `dotnet format` clean, and only `HealthScoreServiceTests.cs` and `DashboardViewModelTests.cs` reference any of the changed members.

PR #2131, opened 2026-09-07
ci: run the integration suite as a non-blocking job so its assertions can fail out loud

### LINKED ISSUE #2101: [Bug]: Nine integration assertions are permanently red and one asserts the wrong last tab — CI only compiles that project
## Problem

`SysManager.IntegrationTests` is compile-checked in CI but never executed (`ci.yml`: "Build integration tests (compile check)"), and the main workstation cannot run `dotnet test`. So a stale expectation there is invisible indefinitely. Two independent rots have accumulated:

### 1. The StatusColors migration never reached this project — 9 red assertions

`StatusColors` members are now theme brush KEYS, not hex strings: `Good = "Success"`, `Neutral = "TextMuted"`, `Bad = "Danger"`. `HealthAnalyzer` assigns those keys to `diag.ColorHex`, and `HealthDiagnostic._colorHex` defaults to `StatusColors.Neutral`. The integration tests still compare against the pre-migration hex literals:

- `HealthAnalyzerExtendedTests.cs:75` — `Assert.Equal("#06D6A0", diag.ColorHex)`, now `"Success"`
- `HealthAnalyzerExtendedTests.cs:237-243` — `AllVerdicts_ProduceValidColor`, 7 `[InlineData]` rows of hex
- `HealthDiagnosticTests.cs:19` — `Assert.Equal("#9AA0A6", d.ColorHex)`, now `"TextMuted"`

Nine failing cases. The other hex literals in the project (`ConvertersTests`, `PingTargetTests`, `DiskHealthReportTests`) pass their own literal in and read it back, so they are unaffected.

### 2. `NavItems_CorrectOrder_AboutLast` asserts a tab that is no longer last

`DeepCleanupViewUiTests.cs:65` asserts `vm.NavItems.Last().Id == "nav-about"`. `grp-advanced` was added after `grp-info`, so the real last item is `nav-env-variables`. The test's intent — About sits at the end of the sidebar — stopped being true when Advanced was appended, and nobody had to decide whether that was intended.

## Expected behavior

Assertions compare against `StatusColors` members rather than duplicating their values, so the next rename cannot rot them silently. The nav-order test asserts what the sidebar is actually supposed to end with, or is rewritten to state the invariant it really means (About is the last entry *of its group*).

Beyond the two fixes, the shape of the problem needs a mechanical answer: a project whose assertions cannot run anywhere in the pipeline will rot again. Either run it somewhere (a non-blocking scheduled job, or a machine that runs the suites as a documented gate) or move the pure assertions into the blocking unit suite.

## Evidence

- `.github/workflows/ci.yml` — integration project is `dotnet build`, never `dotnet test`
- `SysManager/SysManager/Helpers/StatusColors.cs` — `Good = "Success"`, `Neutral = "TextMuted"`
- `SysManager/SysManager/Models/HealthDiagnostic.cs:32` — `_colorHex = StatusColors.Neutral`
- `SysManager/SysManager/ViewModels/MainWindowViewModel.cs` — `grp-advanced` follows `grp-info`

## Affected area

Test suite / CI

laurentiu021:
#2101's two concrete rots were both **already fixed** when I re-derived them against current source. What remained is the part that matters most: the structural reason they were able to rot at all.

| #2101 item | State when checked |
|---|---|
| Nine assertions comparing against pre-migration hex literals | already fixed — all nine reference `StatusColors.Good` / `.Bad` / `.Warning` / `.Info` / `.Neutral` |
| `NavItems_CorrectOrder_AboutLast` asserting the wrong last tab | already fixed — rewritten as `NavItems_AboutIsLastInTheInfoGroup`, an invariant an appended group cannot rot |
| "a project whose assertions cannot run anywhere in the pipeline will rot again" | **open — this PR** |

The 8 hex literals left in the project are the pass-in-and-read-back kind the issue itself called unaffected, and `ci.yml:311` still only compile-checks the project. Both re-verified here rather than taken from the issue text.

## The change

A separate `integration-tests` job, shaped exactly like the `ui-tests` job above it:

- `needs: build-and-test`, `windows-latest`, 30-minute ceiling
- `continue-on-error: true` on the `dotnet test` step
- a warning annotation keyed on **`outcome`**, not `conclusion` — `continue-on-error` forces the step's `conclusion` to `success`, so a conclusion check never fires and the failure is masked completely. The same trap is already documented on `ui-tests`.
- results uploaded as an artifact, 14-day retention
- skipped on fork PRs, which have no secrets and no reason to run a live-system suite

**Non-blocking is deliberate.** These tests touch WMI, the filesystem and real processes, so a runner difference must not gate a merge — and a job that blocks on environment noise gets ignored, which is the failure mode being fixed, not a cure for it. What the job must do is *speak*: the annotation is the point, and the session-start health check reads recent runs, so a red surfaces without anyone remembering to look.

No `junit` logger on this step: that package is referenced only by `SysManager.Tests`, and nothing consumes junit output from this suite. Asking for it would fail the step for the wrong reason and emit a permanent false warning — checked rather than assumed.

## Why this option

The issue offered two: run it somewhere, or move the pure assertions into the blocking unit suite. This is the broader one — it covers every test in the project immediately, including the ones that genuinely need a live system and could never move. Migrating the pure assertions is still worth doing and does not conflict; it is a `test:`-scoped follow-up, not a substitute, and I have not folded it in here to keep this diff to the mechanism.

I cannot predict whether the suite is green on a GitHub runner — this workstation cannot run `dotnet test`, so the first run of this job is also the measurement. That is precisely why it lands non-blocking rather than as a gate.

## Docs

`TESTING.md` and `CONTRIBUTING.md` said "local only, not CI" in three places. Corrected, plus an explicit statement of where a new test belongs: live-system in the integration project, pure in `SysManager.Tests` where it gates merges.

## Verification

- All five workflow files parse; `integration-tests` is registered with the intended `needs`, timeout, seven steps and `continue-on-error: true` — asserted by parsing the YAML, not by reading the diff.
- Actions are SHA-pinned, reusing the exact pins already in this file.
- No code change, so `ci:` — no version bump, no release.
- Leak scan over the staged diff against all 43 current patterns: 0 hits.

Closes #2101

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -490,3 +490,84 @@ jobs:
           name: ui-test-results
           path: TestResults/**
           retention-days: 14
+
+  # The integration suite was compile-checked and never EXECUTED anywhere in the pipeline, and the
+  # primary workstation cannot run `dotnet test` at all. So a stale expectation in it stayed invisible
+  # indefinitely: nine assertions sat red through a whole rename because compiling them still worked,
+  # and one asserted a sidebar tab that had stopped being last. Both were found by reading the file,
+  # which is not a process.
+  #
+  # Non-blocking, deliberately, and shaped exactly like ui-tests above: these tests touch the live
+  # system (WMI, the filesystem, real processes), so a runner difference must not gate a merge. What
+  # they must do is SPEAK — the warning annotation below is the whole point of the job, and the
+  # session-start health check reads recent runs, so a red surfaces without anyone remembering to look.
+  #
+  # It spoke immediately. The first run found five failures and a hang, none of which anything in the
+  # pipeline could have reported before, because the suite was compile-checked and never executed:
+  #   WindowsUpdateAutoCheckTests.CommandExists(propName: "ListFeatureUpdatesCommand")
+  #   EventLogServiceTests.Read_SeverityFilter_ReturnsOnlyRequested
+  #   EventLogServiceTests.Read_System_ReturnsSomeEntries_Within_ShortWindow
+  #   EventLogServiceTests.Read_EntriesEnrichedWithExplanation
+  #   DeepCleanupViewUiTests.View_BindsToDeepCleanupViewModel  (Cannot find resource named 'Display')
+  # then silence from 00:07:41 to the 30-minute ceiling. So this job starts red on purpose: the failures
+  # are real and are being fixed, and a job that hid them would be the thing worth deleting.
+  #
+  # This is the broader of the two answers the issue proposed. Moving the project's pure assertions
+  # into the blocking unit suite is still worth doing and does not conflict; it just cannot cover the
+  # tests that genuinely need a live system, which this does.
+  integration-tests:
+    name: Integration tests (non-blocking)
+    runs-on: windows-latest
+    needs: build-and-test
+    timeout-minutes: 30
+    # Fork PRs have no secrets and no reason to run a live-system suite.
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Setup .NET 10 SDK
+        uses: actions/setup-dotnet@a98b56852c35b8e3190ac28c8c2271da59106c68 # v6.0.0
+        with:
+          dotnet-version: 10.0.x
+
+      - name: Restore
+        run: |
+          dotnet restore SysManager/SysManager/SysManager.csproj
+          dotnet restore SysManager/SysManager.IntegrationTests/SysManager.IntegrationTests.csproj
+
+      - name: Build integration tests
+        run: dotnet build SysManager/SysManager.IntegrationTests/SysManager.IntegrationTests.csproj -c Release --no-restore
+
+      # --blame-hang is not optional here, it is the whole reason this job is usable. The first run of this
+      # job proved the suite HANGS on a clean runner: it reported five failures in the first eight minutes and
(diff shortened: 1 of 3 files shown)


laurentiu021:
## The job spoke on its first run, and what it said matters

The integration suite **hangs** on a clean runner, and had five real failures before it did. None of this could have been reported before, because the project was compile-checked and never executed — which is exactly the structural problem #2101 raised.

```
[00:00:04] WindowsUpdateAutoCheckTests.CommandExists(propName: "ListFeatureUpdatesCommand")  FAIL
[00:05:52] EventLogServiceTests.Read_SeverityFilter_ReturnsOnlyRequested                     FAIL
[00:06:44] EventLogServiceTests.Read_System_ReturnsSomeEntries_Within_ShortWindow             FAIL
[00:07:33] EventLogServiceTests.Read_EntriesEnrichedWithExplanation                          FAIL
[00:07:41] DeepCleanupViewUiTests.View_BindsToDeepCleanupViewModel                           FAIL
           Cannot find resource named 'Display'
[00:07:41 → 00:28]  no output at all, then the 30-minute ceiling cancelled the job
```

### One change to the job as a result

A cancelled job reports `cancelled`, which `gh pr checks` renders as `fail` — indistinguishable from a broken pipeline, and **permanently** red. That is precisely how a non-blocking job teaches everyone to ignore it, which would have made this worse than no job at all.

So the run now carries `--blame-hang --blame-hang-timeout 5m`. It stops after five minutes of silence, **names** the test that stopped responding, and writes a dump beside the results. A hang becomes a finding instead of a timeout.

### This job starts red on purpose

The five failures are real. They are being fixed next, and this job is what verifies the fixes — I cannot run the suite on this workstation, which is the same reason it rotted in the first place. A job that hid them would be the thing worth deleting.

The workflow comment records all five inline, so the next reader does not have to find this thread.

Issue #1633, opened 2026-08-03
[Enhancement]: Cross-app - Corner radii bypass the RadiusSm/Md/Lg token scale in views: 177 raw values vs 1 token use


laurentiu021:
## Problem
The radius tokens (`RadiusSm`=4, `RadiusMd`=8, `RadiusLg`=12, `RadiusXl`=16, `RadiusPill`=999, App.xaml:132-136) are consumed 22 times inside App.xaml's own control styles - that migration worked. But it stopped at the App.xaml boundary. Across Views/*.xaml + MainWindow.xaml there are 177 raw numeric `CornerRadius` values and exactly 1 token reference (Views/LogsView.xaml:14). 124 of those raw values are literally the token numbers written out by hand (`CornerRadius="8"` x61, `"12"` x46, `"4"` x8, `"16"` x1, `"999"` x1), so the scale is being obeyed by convention while the token is bypassed. Three values sit off-scale: `CornerRadius="2"` (x49, the admin-banner accent stripe), `"45"` (DashboardView.xaml:91, a 90x90 circle) and MainWindow.xaml:298 `"999"`.

## Proposed solution
Migrate the 124 on-scale raw values in Views/ + MainWindow.xaml to `{StaticResource RadiusSm/Md/Lg/Xl/Pill}`, matching what App.xaml already does. Add one missing token for the genuinely-used stripe radius - `<CornerRadius x:Key="RadiusStripe">2</CornerRadius>` - since 49 uses is not noise and it is the accent stripe shared by the admin banner and DevelopmentBanner. Leave DashboardView.xaml:91 `CornerRadius="45"` raw with a comment: it is a geometric half-of-90 circle, not a scale value. Normalise MainWindow.xaml:298 `"999"` to `RadiusPill`. Because the radius tokens are `StaticResource` and preset-invariant by design (App.xaml:127-131), this is a zero-runtime-cost, zero-visual-change swap for the 124.

## Rationale
Unlike the spacing tokens, radius tokens are already proven in this codebase and already fix a documented past problem - the token commit's CHANGELOG says the scale killed "per-style radius drift (was 5/6/8/10/999 scattered)". Views were simply left out of that fix, so the drift can silently return there: nothing stops the next view from using 10 or 6, and no check would catch it. This is the highest-certainty, lowest-risk consistency win available in the design system: the values are already correct, only the reference is missing, so it is mechanical and visually inert.

## Evidence
[...]

Issue #1618, opened 2026-08-03
[Enhancement]: Cross-app - Extract the copy-pasted admin elevation banner pair into a shared AdminBanner control (30 views, 60 bl...


laurentiu021:
## Problem
The golden admin-control contract is hand-copied into every privileged view instead of living in one reusable control. Each of 30 views carries TWO near-identical `<Border>` blocks - a "not elevated" one and an "elevated" one - differing only in the sentence text.

Not-elevated banner (30 copies): `Background="{DynamicResource Surface2}" BorderBrush="{DynamicResource Border1}" BorderThickness="1" CornerRadius="8" Padding="14,12"` + a Segoe Fluent `&#xE83D;` glyph + an `AdminButton`. Verified byte-identical brushes/radius/padding in all 30 (only Margin differs by layout strategy). Example: Views/CleanupView.xaml:23-34, Views/StartupView.xaml:29-43.

Elevated banner (30 copies): `Background="{DynamicResource WarningBgSubtle}" BorderBrush="{DynamicResource WarningBg}" CornerRadius="12" Padding="12,8"` + a nested 4px `WarningStripe` stripe `Border` with the magic `Margin="-12,-8,0,-8"` negative-padding trick. Example: Views/CleanupView.xaml:39-51, Views/ServicesView.xaml:46-58.

The drift this already produced is measurable: the elevated banner's icon is the astral emoji `&#x1F6E1;` with NO FontFamily in 27 views (Views/CleanupView.xaml:46, Views/PrivacyView.xaml:53, +25 more) but the Segoe Fluent `&#xE83D;` in 3 views (Views/ServicesView.xaml:53, Views/StartupView.xaml:53, Views/ProcessManagerView.xaml). Those same 3 views are also the only ones that give the icon `Foreground="{DynamicResource WarningText}"`; the other 27 leave it default-coloured.

## Proposed solution
[...]

PR #2136, opened 2026-09-07
fix: three corner radii were off the scale, and nothing would have caught a fourth

laurentiu021:
The radius tokens (`RadiusSm/Md/Lg/Xl/Pill`) exist because the radii drifted once already — the commit that added them says the scale killed *"per-style radius drift (was 5/6/8/10/999 scattered)"*. App.xaml adopted them. **The views and the shell did not**, so the drift could return there unnoticed. It had.

## The seven off-scale values

All found by counting, not by eye.

| Value | Where | What it is |
|---|---|---|
| `10` | MainWindow's update banner | drift — every neighbouring surface is 12 |
| `10` | MainWindow's success card | drift — same |
| `10` | Tune-Up verdict badge, Landing tab | drift |
| `9` | toggle thumb | half of its own 18px height |
| `11` | toggle track | half of its own 22px height |
| `14` | shell icon disc | half of its own 28px |
| `45` | health-score disc | half of its own 90px |

**The three 10s are corrected** to the nearest scale step: `RadiusLg` for the two surfaces whose neighbours are 12, `RadiusMd` for the badge — 12 would clamp to nearly a pill on a 19px-tall element. Those are 2px visible changes, and the reason this is `fix:` rather than `refactor:`.

**The other four are a pill or a circle written as a number.** They now say `RadiusPill`, which WPF clamps to exactly the same pixels — no visual change, the intent is stated, and the arithmetic cannot go stale if the element is ever resized.

## The guard asks for a value, not a token

`EveryCornerRadius_IsOnTheScale` requires an **on-scale value**. Requiring a *token* across 175 call sites is a separate, purely mechanical migration; requiring the value is what prevents the regression, and it is checkable without touching a single view.

`2` is on the allowed list because it is the accent stripe on the elevation and preview banners — a deliberate hairline, not a surface radius.

## Three corrections to #1633

Each re-derived from source.

| The issue says | Current source says |
|---|---|
| `CornerRadius="8"` ×61, `"12"` ×46 | **39 and 77** |
| Three values are off-scale: `2`, `45`, `999` | It missed **`10` ×3, `9`, `11`, `14`** — the four that are actually drift, and the three that are actually *wrong* |
| Add a `RadiusStripe` token, because "49 uses is not noise" | Invalidated by #1618: the stripe now lives in `AdminBanner`, `DevelopmentBanner` and Uninstaller — about **three** uses. Not worth a token; the allowed-value list carries it |

## Deliberately not here

The bulk migration of the 175 on-scale literals to tokens. It is visually inert, touches 63 files, and the issue itself rates it **value 2/5**. It belongs in its own `refactor:` PR after #2134 lands, so the two do not collide across 30 shared view files. That is why this closes nothing and only refs #1633.

## Red/green — 6 mutations, all as claimed

| # | Mutation | Expected | Result |
|---|---|---|---|
| 1 | Restore one `10` | names the file and the value | RED, both |
| 2 | Restore all three `10`s | names all three, across two files | RED, 3/3 |
| 3 | Restore the `45` | named too — geometry is not special-cased | RED |
| 4 | Add a plausible-but-off-scale `6` | named — the rule is the **scale**, not a list of past mistakes | RED |
| 5 | Add an on-scale `16` | **stays green** — not an allowlist of what the tree happens to contain | GREEN |
| 6 | Break the pattern | vacuity floor fires, not every value passing | RED, floor |

Mutations 4 and 5 are the pair that matters: together they show the guard encodes the scale rather than the current contents of the tree.

## Verification

- Four projects build in Release: 0 errors, 0 warnings.
- `dotnet format --verify-no-changes`: clean.
- Guard sweep over 107 names: **113 green, 1 red** — the throwaway local runner's own `Program.cs`, not in the repo.
- Leak scan over the staged diff against all 43 current patterns: 0 hits.

Deferred to the secondary workstation: the three 2px changes seen side by side with their neighbours, and the four pill/circle conversions confirmed pixel-identical.

## Ordering note

This, #2134 and #2135 all claim 1.76.26, because CI checks the version against a bump from the newest tag (`v1.76.25`) and there is only one next number. Whichever merges second and third get rebased and renumbered; they are independent and none blocks the others.

Refs #1633

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,25 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.76.28] - 2026-09-07
+
+Three rounded corners were slightly the wrong roundness, which is the kind of thing nobody notices one at a
+time and everybody notices in aggregate: a panel that looks almost, but not quite, like the panel above it.
+
+### Fixed
+- **The update notice and the "all clear" card now have the same corners as every other panel.** Both were
+  2px squarer than their neighbours. Same for the small verdict badge on the Landing tab after a Tune-Up.
+
+### Changed
+- **Four circles and pills now say they are circles and pills.** Their roundness was written as a number that
+  happened to be half the element's size — correct today, quietly wrong the moment the element is resized.
+  No visible change: the app rounds them to exactly the same pixels.
+
+### Added
+- **A check that a corner radius has to be one of the app's own five sizes.** These sizes exist because the
+  corners drifted once before; the check covers the places that were left out of that fix, so they cannot
+  drift again.
+
 ## [1.76.27] - 2026-09-07
 
 Four result messages were being cut off at the edge of their box instead of continuing onto a second line.
--- a/SysManager/SysManager.Tests/ArchitectureTests.cs
+++ b/SysManager/SysManager.Tests/ArchitectureTests.cs
@@ -2141,6 +2141,72 @@ public partial class ArchitectureTests
         return end.Success ? rest[..end.Index] : rest;
     }
 
+    /// <summary>
+    /// Every corner radius is either a token or one of the scale's own numbers. Nothing in between.
+    /// </summary>
+    /// <remarks>
+    /// The radius tokens exist because the radii had already drifted once — the commit that added them says
+    /// the scale killed "per-style radius drift (was 5/6/8/10/999 scattered)". App.xaml adopted them; the
+    /// views and the shell did not, so the drift could come back there unnoticed, and it had:
+    /// <list type="bullet">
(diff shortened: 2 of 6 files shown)


laurentiu021:
Rebased onto the merged banner extraction (#2134). Two things changed in the rebase, both worth flagging:

**The vacuity floor had to be re-measured, from 150 against 177 to 80 against 90.** The extraction removed 60 banner Borders and 27 of the stripe hairlines, so the population of numeric radii fell by nearly half and the floor fired on the rebase. That is the floor working: a population is a measurement, and another change moved it. Current spread is `8` ×39, `2` ×20, `12` ×19, `4` ×9, `999` ×2, `16` ×1 — the allowed set is unchanged and still exactly right.

**The version is now 1.76.27**, since #2134 took 1.76.26.

Re-verified after the rebase: 4 projects build 0/0, `dotnet format` clean, guard sweep 108 green / 1 red (the throwaway local runner's own `Program.cs`).

Issue #2141, opened 2026-09-07
[Bug]: Integration test run aborts partway, and blame names the same test both times


laurentiu021:
## Problem

The integration test host runs every test, reports, and then never exits. `--blame-hang` collects a dump five minutes later and the job fails on the timeout rather than on a test.

From the first execution of the suite ([run 34142338870](https://github.com/laurentiu021/SystemManager/actions/runs/34142338870)):

```
16:22:31  dotnet test ... --blame-hang --blame-hang-timeout 5m
16:30:07  last test reports
16:35:22  Data collector 'Blame' message: The specified inactivity time of 5 minutes has elapsed.
          Collecting hang dumps from testhost and its child processes.
16:35:26  Failed!  - Failed: 7, Passed: 554, Skipped: 0, Total: 561, Duration: 7 m 32 s
```

All 561 tests ran and reported inside 7 m 32 s. The five minutes after that are pure idle: nothing is executing, and the process simply does not shut down. So this is a **shutdown** hang, not a mid-run one — the earlier reading of it as "the suite hangs partway" was wrong.

Something the suite creates keeps the process alive after the last test. Candidates worth checking, in the order they are cheap to check:

- A foreground (non-background) thread. `StaHelper` sets `IsBackground = true`, so its STA threads are not it, but anything else that spawns a thread may not.
- The WPF `Application` the view tests create. It is constructed with `ShutdownMode.OnExplicitShutdown` and never shut down, and its `Dispatcher` belongs to an STA thread that has since exited.
- A service started by a constructed view model that owns a timer, watcher, or poll loop and is never disposed, since the tests construct real services rather than substitutes.

## Steps to reproduce

Run the integration job. It reproduces on every execution so far.

## Expected behavior

The test host exits promptly after the last test reports, and the job's result reflects the tests rather than the inactivity timeout.

## Affected tab

None — this is the test suite, not the product. Filed as a bug because it makes the integration job's result unreadable: a real failure and a hang produce the same red.

## Notes

The hang dump is uploaded as a job artifact, which should name the threads still alive and settle which of the candidates above it is. Worth reading that before changing anything.

Found by #2131, which ran this suite for the first time. The seven test failures it also exposed are repaired in #2139; this is deliberately not bundled with them, because it is one problem with one cause and the repairs are seven unrelated ones.

laurentiu021:
## Correction: this is not a shutdown hang, and the run is being aborted

[...]

PR #2146, opened 2026-09-07
fix: Cancel could not stop a severity-filtered System Logs read for minutes

### LINKED ISSUE #2140: [Bug]: System Logs — Cancel does not stop a filtered read for minutes
## Problem

Cancelling a System Logs read does not stop it. Pressing Cancel while a filtered query is running has no effect until the underlying read finishes on its own, which can take minutes — the tab stays busy, Refresh stays disabled, and nothing tells the user why.

The token is observed in the right places but cannot reach the operation that is actually blocking. In `EventLogService.Enumerate`:

```csharp
while (!ct.IsCancellationRequested && emitted < opt.MaxResults)
{
    ...
    try { rec = await Task.Run(readNext, ct).ConfigureAwait(false); }
```

The `ct` passed to `Task.Run` prevents the delegate from **starting**. Once `EventLogReader.ReadEvent()` is running, nothing can interrupt it: the OS walks records inside `EvtNext` looking for one that satisfies the XPath, and returns only when it finds a match or exhausts the result set. So a query whose filter matches nothing over a long window blocks in a single call, and the loop's `ct.IsCancellationRequested` check is not reached until after that call returns.

The loop is correct about everything else — it cancels between records and does not swallow the cancellation. The gap is inside one record read.

## Evidence

Measured on a CI runner, `EventLogServiceTests.Read_SeverityFilter_ReturnsOnlyRequested`:

- query: `System`, 90-day window, `MaxResults = 20`, severities `Error` + `Critical`
- budget: a 10-second `CancellationTokenSource`
- observed: **4 m 36 s**, then `OperationCanceledException`

Its two siblings over the same log with **no severity filter** took 20 s and 26 s. The filter is what makes the first read expensive: with few matching events in the window, one `ReadEvent()` scans nearly the whole log.

The same three tests fill their caps in well under a second on a developer machine, which is why this was never noticed: the blocking read only becomes visible when the filter cannot be satisfied quickly.

## Steps to reproduce

1. Open System Logs on a machine with few or no Error/Critical events in the last 90 days.
2. Set the window to 90 days and filter to Error + Critical.
3. Start the read, then press Cancel.
4. The tab stays busy for as long as the underlying scan takes. Cancel appears dead.

## Expected behavior

Cancel stops the read within a second or so, the tab leaves its busy state, and partial results are handled by the existing cancel branch.

## Affected tab

System Logs.

## Notes on a fix

`EventLogReader.ReadEvent(TimeSpan timeout)` returns `null` on timeout, which would let the loop poll in bounded slices and observe the token between them. The obstacle is that `null` currently means "end of the result set" (`if (rec is null) yield break;`), so a timeout and an exhausted query would be indistinguishable — the two cases need separating before that overload can be used. That is a design decision, not a mechanical change, which is why this is filed rather than patched inside the test-repair PR that found it.

Whatever the shape, the fix needs a test that pins it: `Enumerate` already takes a caller-supplied `Func<EventRecord?> readNext` precisely so the cancellation path can be driven without a real reader, so a slow stub can prove cancel-during-a-read deterministically.

laurentiu021:
The severity filter was a `Level` clause in the XPath the Event Log service evaluates itself. That means a single `EventLogReader.ReadEvent()` walks records internally until one satisfies the query and returns only then — and it is a blocking native call no `CancellationToken` can interrupt.

On a window containing few matches, one read blocked for **4 m 43 s against a 10-second token** on a CI runner. For that whole time Cancel did nothing, the tab stayed busy, and Refresh stayed disabled because it is gated on not-busy.

The loop around it was already right: it checks the token between records and does not swallow the cancellation. The gap was entirely inside one record read, where the token cannot reach.

## The filter moves into managed code

`BuildXPath` keeps `Since`, `Provider` and `EventID` — bounds the OS can apply cheaply — and drops `Level`. Each `ReadEvent()` now returns the next record in the window immediately, which makes the loop's existing `ct.IsCancellationRequested` check the cancellation point.

Three things about the trade, because it reads like a pessimisation and is not:

- An **unfiltered** read is unchanged. Every record it reads is emitted either way, and that is the Logs tab's default.
- A filtered read examines more records but does not do more **total work**. The OS was already walking them; it just did so uninterruptibly and without telling anyone.
- `Matches()` is consulted **before** `Project()`, because `Project` formats the record's message — the expensive part — and a filtered read discards most of what it reads. Reading `rec.Level` is cheap.

`Matches()` uses the same `MapLevel` the projection uses, so a record cannot be admitted by the filter and then labelled with a different severity in the list.

## `SeverityToLevels` is deleted, and its invariant is not

It existed only to build that clause. With the clause gone it was production code kept alive by its own test.

What it guarded is kept: `MapLevel` folds Level 0 (LogAlways) into Info, so an Info filter must admit **both** 0 and 4 — now asserted against `Matches`, the code that actually does the filtering. Three `BuildXPath` tests that asserted the `Level` clause are migrated the same way rather than deleted, and a new one asserts the clause is **absent**, because the only thing stopping it being reinstated later as an obvious-looking optimisation is a test that fails when it comes back.

## The regression test, and why it filters on Verbose

`Read_Cancellation_WithASeverityFilter_StopsFast` asks for Verbose, which the System log essentially never contains. The defect only appears when the filter **cannot** be satisfied quickly: filtering on Error would fill its cap in milliseconds and pass against the very code it exists to catch.

## Red/green

| | mutation | result |
|---|---|---|
| | baseline | unit 11 green 0 red · integration 2 green 0 red |
| Q1 | restore the `Level` clause | **RED** the absence guard fires |
| Q2 | `Matches` → always true | **RED** 4 unit reds + `Read_SeverityFilter` |
| Q3 | `MapLevel` → unknown/0 folds to Verbose | **RED** the LogAlways row, exactly as its ancestor guarded |
| Q4 | `Enumerate` → drop the `Matches` guard | **RED** `Unexpected severity Info on entry 1` |
| | restored, sha verified, rebuilt | unit 11 green 0 red · integration 2 green 0 red |

**Q1 needs a caveat rather than a claim.** It reds the XPath absence guard, but it does **not** red the cancellation test on this workstation — and that is measured, not assumed. A System log of 28,997 records answers the same unsatisfiable `Level=5` query in **333 ms**, so the blocking read finishes inside the test's budget. The 4 m 43 s came from a CI runner, so the integration job is where that test can go red.

The first attempt at Q1 was also wrong in a way worth recording: it mapped severity to Level by enum ordinal, producing `Level=4 or Level=3` for Info+Error — a query matching plenty of events, and therefore proving nothing. The faithful mapping is the one the deleted `SeverityToLevels` carried.

## Verification

- Four projects build **0 errors, 0 warnings**.
- **107 of 108** guards green. The single red is `EverySourceFile_CarriesTheAuthorHeader` flagging the throwaway local runner, which is not in the repository.
- The 29 `EventLogService` unit tests: **43 green** (theories expanded), 0 red.
- The 7 integration reads: green, in **0.5 s** against a 0.47 s baseline measured before the change — so no local regression from client-side filtering.
- `dotnet format --verify-no-changes` clean; leak scan 43 patterns over 5 files, its one hit the Windows assistant privacy toggle in a historical changelog line.

The observable proof lands after merge: `Read_SeverityFilter_ReturnsOnlyRequested` took 4 m 43 s in the integration job and should now be comparable to its unfiltered siblings.

Closes #2140.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,23 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.77.2] - 2026-09-07
+
+Cancel now works in System Logs. Filtering the log to Errors and Criticals over a long period could leave the
+tab stuck for minutes with Cancel doing nothing at all — four and a half minutes on one slower machine, against
+a stop asked for after ten seconds.
+
+### Fixed
+- **Pressing Cancel while System Logs is reading now stops it.** The filtering was being done by Windows
+  itself, inside a single call SysManager had no way to interrupt: it searched the whole period for a matching
+  event before handing anything back, and until it did, Cancel and Refresh could do nothing. SysManager now
+  reads the events and applies the severity filter itself, so it can stop between one event and the next. An
+  unfiltered read — what the tab does by default — behaves exactly as before.
+
+### Added
+- **A test that a cancelled filtered read comes back quickly.** It asks for a severity the log almost never
+  contains, which is the case that used to hang: there is nothing to find, so the search ran to the end.
+
 ## [1.77.1] - 2026-09-07
 
 If you use a screen reader, every tab has been silent about what it was doing. Start a system repair or a
@@ -31,6 +48,7 @@ on screen and none of it out loud. Fifty-two tabs now report themselves as they 
 - **A check that every status line stays announced, and that the fast-moving numbers stay quiet.** Both halves
   matter. A well-meant change that announced everything would look like an improvement and would make the app
   unusable with a screen reader, so the quiet ones are named and pinned too.
+
 
 ## [1.77.0] - 2026-09-07
 
--- a/SysManager/SysManager.IntegrationTests/EventLogServiceTests.cs
+++ b/SysManager/SysManager.IntegrationTests/EventLogServiceTests.cs
@@ -114,6 +114,59 @@ public class EventLogServiceTests
     }
 
     /// <summary>
+    /// Cancelling a SEVERITY-FILTERED read stops it, which is the case that used to hang.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="Read_Cancellation_StopsFast"/> above covers an unfiltered read and always passed. The
+    /// filtered one did not: the severity filter used to be a <c>Level</c> clause in the XPath, which the
+    /// Event Log service evaluates itself, so a single <c>ReadEvent()</c> walked records internally until it
+    /// found a match and returned only then. That native call cannot be interrupted by a token, so on a
+    /// machine with few Error or Critical events in the window the read blocked for minutes with Cancel doing
+    /// nothing — measured at 4 m 43 s against a 10-second budget on a CI runner.
+    /// <para>The window is wide, the cap unreachable, and the severity deliberately SCARCE, because the
+    /// defect only appears when the filter cannot be satisfied quickly. Filtering on Error would pass either
+    /// way on a machine with plenty of errors — the cap fills in milliseconds and nothing ever blocks — which
+    /// would make this green against the very code it exists to catch. Verbose (Level 5) is essentially
+    /// absent from the System log, so an OS-evaluated query for it has to walk the whole window.</para>
+    /// <para><b>It still only fails where the scan is slow, and that is worth being exact about.</b> Restoring
+    /// the old <c>Level</c> clause and running this on a developer machine leaves it GREEN: a System log of
+    /// 28,997 records answers the same unsatisfiable query in 333 ms, comfortably inside the budget. The 4 m
+    /// 43 s came from a CI runner, so that is where this test can go red, and the integration job is where it
+    /// runs. A test that can only fail on some machines is a weak test; it is kept because the alternative is
+    /// no regression test at all for a defect that was measured, and because a machine fast enough to pass it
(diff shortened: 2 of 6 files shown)


laurentiu021:
## The integration job found a fourth copy of the stale assertion

First run of this branch after #2131 landed on main, and it caught something the local sweep did not:

```
Failed SysManager.IntegrationTests.QaAuditTests.EventLogService_BuildXPath_CombinesAllFilters [2 ms]
Failed!  - Failed: 1, Passed: 424, Total: 425
```

`QaAuditTests` reaches `BuildXPath` **by reflection, from a different project**, so grepping the method name across the unit tests found the three copies in `SysManager.Tests` and missed this one. Fixed in `18b26df`, migrated identically: the `Level=2` assertion becomes `DoesNotContain("Level")`, so the clause cannot be reinstated later without a test failing.

Swept all three test projects afterwards for any remaining `Level=` assertion or `BuildXPath` reflection site — none left. The other two reflection sites in that file are about provider-name quoting and the no-filter case, and neither touches severity.

Worth stating plainly: I claimed a propagate sweep in the commit message and it was incomplete, because it keyed on the deleted method's name rather than on the behaviour. A reflection call from another assembly is invisible to that kind of grep. The job is what closed the gap.

## Also visible in the same run: the #2139 repairs took

That run reported **1 failure out of 425**, against **7 out of 561** on the first execution of this suite. The six repaired in #2139 are gone, and the only red was the one this branch introduced.

The run still aborts partway — #2141, unchanged and expected. This is now the third execution with a different total (561, 215, 425), which is more evidence that it stops at a variable point rather than finishing.

Issue #2149, opened 2026-09-07
[Bug]: Elevated PowerShell leaks a runspace per call, and opening one can hang forever


laurentiu021:
## Problem

When SysManager runs elevated, every PowerShell invocation creates an out-of-process runspace and a `powershell.exe` 5.1 child process, and the teardown does not release them. Separately, the call that opens that runspace can never return, and the code has no answer for that case.

Both were found in a hang dump from the integration job (#2141), which is the only place the volume gets high enough to make them visible. The analysis is [here](https://github.com/laurentiu021/SystemManager/issues/2141#issuecomment-5575028988); the measurements below come from it.

Only the elevated path is affected. `CreateRunspace` branches on it:

```csharp
if (!_isElevated)
{
    var initialSessionState = InitialSessionState.CreateDefault2();
    ...
    return (RunspaceFactory.CreateRunspace(initialSessionState), null, null);   // in-process, no child
}

var processInstance = new PowerShellProcessInstance(new Version(5, 1), ...);
runspace = RunspaceFactory.CreateOutOfProcessRunspace(TypeTable.LoadDefaultTypeFiles(), processInstance);
```

### (a) The runspace teardown leaves its transport threads alive

`DisposeRunspaceResources` disposes and never closes:

```csharp
try { runspace?.Dispose(); }
finally { try { processInstance?.Dispose(); } finally { process?.Dispose(); } }
```

In the dump, 1,112 of 1,447 managed threads were PowerShell remoting transport threads — roughly four per runspace, so about **276 out-of-process runspaces still alive**, each with a child process behind it:

```
560  ClientSessionTransportManagerBase.ProcessMessageProc -> SemaphoreSlim.WaitCore
276  OutOfProcessClientSessionTransportManager.ProcessOutputData -> StreamReader.ReadLine -> ReadFile
276  OutOfProcessClientSessionTransportManager.ProcessErrorData -> same
```

There is an ordering hazard as well: 275 of those runspaces were stuck mid-`Open()`, and disposing a runspace that never finished opening is the case least likely to tear down cleanly.

### (b) `RemoteRunspace.Open()` has no timeout, so a wedged host hangs instead of failing

This is the half that matters to a user. 275 threads in the dump were blocked here:

```
System.Management.Automation.RemoteRunspace.Open
System.Management.Automation.Runspaces.AsyncResult.EndInvoke
System.Threading.WaitHandle.WaitMultiple
```

[...]

PR #2151, opened 2026-09-08
fix: the elevated runspace's powershell.exe outlived it, and every leftover held threads open

laurentiu021:
Disposing a `System.Diagnostics.Process` releases the **handle**. It does not stop the process. Teardown did only that:

```csharp
finally { try { processInstance?.Dispose(); } finally { process?.Dispose(); } }
```

So when the `powershell.exe` 5.1 child behind an out-of-process runspace did not exit on its own, it stayed — and its stdout and stderr pipes stayed open with it, which kept the remoting transport's reader threads blocked in `ReadFile` forever.

## Measured, not inferred

A hang dump from the integration job showed **1,447 managed threads** in one test host, of which **1,112 were PowerShell remoting transport threads**:

| count | frame |
|---:|---|
| 560 | `ClientSessionTransportManagerBase.ProcessMessageProc` |
| 276 | `OutOfProcessClientSessionTransportManager.ProcessOutputData` |
| 276 | `OutOfProcessClientSessionTransportManager.ProcessErrorData` |
| 275 | `RemoteRunspace.Open` — a host too saturated to complete another handshake |

About four threads per runspace, so roughly **276 children still alive**. #2150 bounded that `Open()` wait; this removes the saturation that caused it.

**Elevated only**, which is why it never showed up locally: `CreateRunspace` takes the out-of-process branch only when `_isElevated`, and a CI runner is elevated while a developer session is not.

## The change

`ReleaseProcess` stops a child that is still running, then disposes the handle. It mirrors `TryTerminateForCancellationAsync`'s handling of the same races — the process can exit between the `HasExited` check and the kill, a tree termination can fail outright — because they are the same races. The difference is that this one has nothing to report to: it runs in a `finally`, so a failure to stop the child leaks what was already leaking, while throwing would additionally lose the runspace's disposal.

**Closing the runspace before disposing it is deliberately not done.** It is the obvious next idea and it carries a real risk: 275 of those runspaces were parked inside `Open()`, and `Close()` on a runspace that never finished opening can block on the same handshake. Stopping the child is what frees the threads, and it cannot block on the runspace's state.

**Reusing one runspace per session** instead of creating one per invocation is the structural fix and is **not here**. It is a lifecycle change in a privileged path and wants its own review; #2149 stays open for it. This makes each invocation clean up after itself, which is what turns 276 leaks into none.

## Red/green

| | mutation | result |
|---|---|---|
| | baseline | unit 2 green · integration 1 green |
| T1 | `ReleaseProcess` → dispose only, do not stop | **RED** the child outlives the run; the bounded wait is cancelled |
| T2 | hand teardown a bare `Dispose` instead | **RED** same, via the production wiring |
| T3 | drop the `HasExited` guard | GREEN, deliberately — the guard narrows a race and is not the thing under test |
| | restored, sha verified, rebuilt | unit 2 green · integration 1 green |

## The test took three attempts, and two of them were false greens

Worth recording, because each attempt found a real flaw in the previous one:

1. **It invoked the private release by reflection.** That proved what the method does and nothing about whether the production path uses it — T2 left it green. It now goes through `RunAsync` with the child injected via the `createRunspace` seam, so wiring and behaviour are one assertion.
2. **The stand-in was `cmd /c pause` with redirected stdin.** Disposing the `Process` closes its redirected handles, `pause` saw EOF, and the child exited on its own — so T1, which removes the termination entirely, *still* left it green. Nothing is redirected now: disposing the object cannot touch the process, and only a kill ends it.
3. **The proof runner itself reported GREEN when the harness produced no summary at all.** T1 had killed it, via an orphaned child writing to a closed pipe. It now treats a missing summary as BROKEN rather than as an absence of reds.

Observed through a **second, independent handle** to the same process, because teardown disposes the one it is given and querying that afterwards throws *"No process is associated with this object"* — which says nothing about whether the process died. The independent handle lets `WaitForExitAsync` be awaited rather than polled, so there is no sleeping and nothing to go flaky.

`DisposeRunspaceResources_ReleasesTheChildThroughTheSuppliedAction` is renamed to `..._ReleasesTheChildLast`, which is what it actually covers. It asserts ordering, not wiring, and the old name claimed otherwise.

## Verification

- Four projects build **0 errors, 0 warnings**.
- The `PowerShellRunner` unit tests: **44 green**.
- **107 of 108** guards green. The single red is `EverySourceFile_CarriesTheAuthorHeader` flagging the throwaway local runner, which is not in the repository.
- `dotnet format --verify-no-changes` clean; leak scan 43 patterns over 5 files, its one hit the Windows assistant privacy toggle in a historical changelog line.

The observable confirmation lands after merge: the integration job's thread count should stop climbing, and #2141's one never-completing test should complete.

Part of #2149; the per-session reuse half stays open there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,24 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.78.2] - 2026-09-08
+
+Running SysManager as administrator left a hidden `powershell.exe` behind every time a feature used one. They
+were invisible, they never went away on their own, and each one held a few threads open inside SysManager. A
+long session with a lot of Windows Update or disk-health checks accumulated them.
+
+### Fixed
+- **The helper SysManager starts for administrator tasks is now actually stopped when the task ends.** It was
+  being let go of rather than closed — the difference being that letting go of something does not end it. Each
+  leftover kept its output pipes open, which kept threads inside SysManager waiting on them for nothing. In a
+  test run that used PowerShell hundreds of times, this left roughly 276 of them alive at once and over a
+  thousand stuck threads, until the run could no longer make progress.
+
+### Added
+- **A test that starts a real background process and proves the cleanup stops it.** Written the awkward way on
+  purpose: the first two versions passed against code that did no cleanup at all, once because the stand-in
+  process was dying for an unrelated reason and once because the test never went through the real code path.
+
 ## [1.78.1] - 2026-09-07
 
 A tab that uses Windows PowerShell can no longer sit "working…" forever. If the PowerShell host on your PC
--- a/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
+++ b/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
@@ -11,6 +11,74 @@ namespace SysManager.IntegrationTests;
 [Collection("Network")]
 public class PowerShellRunnerTests
 {
+    /// <summary>
+    /// A run's teardown STOPS the child process the runspace was built with, rather than only dropping its
+    /// handle.
+    /// </summary>
+    /// <remarks>
+    /// #2149(a), end to end through <c>RunAsync</c>. Disposing a <see cref="System.Diagnostics.Process"/>
+    /// releases the HANDLE and does not stop the process. When <c>powershell.exe</c> outlived its runspace, its
+    /// stdout and stderr pipes stayed open and the remoting transport's reader threads stayed blocked in
+    /// <c>ReadFile</c> forever — a hang dump from the integration job showed 1,112 such threads, about four per
+    /// runspace, so roughly 276 children still alive in a single test host.
+    /// <para><b>Through RunAsync, not by invoking the release directly.</b> An earlier version called the
+    /// private method by reflection, which proved what the method does and nothing about whether the production
+    /// path uses it — a mutation that handed teardown a bare <c>Dispose</c> instead left it green. Injecting the
+    /// child through the <c>createRunspace</c> seam and letting <c>RunAsync</c> tear it down is what closes
+    /// that: it covers the wiring and the behaviour in one assertion.</para>
+    /// <para>An in-process runspace is used, so the run itself needs no elevation; the child injected beside it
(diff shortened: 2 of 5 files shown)


laurentiu021:
## Measured on this PR's integration run: the leak is gone, the hang is not

The body predicted two things. One held, one did not, and the one that did not is worth being explicit about.

**The leak is gone, and by a wide margin.** Comparing hang dumps from the same suite before and after:

| | before | after |
|---|---:|---:|
| managed threads | **1,447** | **44** |
| `ProcessMessageProc` | 560 | 0 |
| `ProcessOutputData` | 276 | 0 |
| `ProcessErrorData` | 276 | 0 |
| `RemoteRunspace.Open` (parked) | 275 | 0 |

Not one PowerShell remoting transport thread survives. So the child process really was what held them open, and stopping it really does release them.

**The hang is not gone.** The run still aborts, with the same one test never completing:

```
Passed!  - Failed: 0, Passed: 201, Total: 201, Duration: 4 m 23 s
Test Run Aborted.
recorded 202, completed 201, not completed: DeepCleanupViewUiTests.NavItems_DeepCleanup_AfterCleanup
```

So the prediction in the body — *"#2141's one never-completing test should complete"* — was wrong. The leak was not the cause of the hang; it was hiding it. With 1,447 threads of noise removed, the 44 that remain are readable, and they name it:

```
System.Windows.Threading.DispatcherOperation.Wait
System.Windows.Threading.Dispatcher.InvokeImpl
System.Windows.Threading.Dispatcher.Invoke(System.Action)
SysManager.ViewModels.StartupViewModel+<ScanAsync>.MoveNext()
```

and the same shape under `ServicesViewModel.RefreshAsync`. Ten threads blocked in `Dispatcher.Invoke`, seven waiting on a `DispatcherOperation`. A **synchronous** marshal to a dispatcher with nothing pumping it — which is what an `Application` created on an STA thread that has since exited leaves behind.

Filed as its own issue with the source trace; the common path is `ToastService.Show`, which does a blocking `dispatcher.Invoke` and is called at the end of a great many async operations.

Worth noting for the record: this is the dead-dispatcher hypothesis I posted on #2141 and then **refuted**, because a minimal local reproduction did not hang. The refutation was sound for what it tested and wrong about the conclusion — the minimal repro constructed a view model but never ran `ScanAsync`, so it never reached a marshal. The dump is what settled it.

Nothing here changes the case for merging: T1 and T2 prove the child is now stopped, and 1,447 → 44 is the measurement. It simply does not close #2141.

Issue #2152, opened 2026-09-08
[Bug]: Ten dispatcher marshals wait for a UI thread that may never run them


laurentiu021:
## Problem

Ten places marshal to the WPF dispatcher with the **synchronous** `Invoke` from background threads that have nothing to gain by waiting. Each one guards on `Application.Current` being non-null, which answers the wrong question: it says an `Application` exists, not that anything is pumping its dispatcher. When nothing is, the guard passes, `Invoke` queues an operation no one will run, and the calling thread waits forever.

This is the remaining cause of the hang in #2141. #2151 removed 1,403 leaked threads of noise from the hang dump, and the 44 that were left named it: ten blocked in `Dispatcher.Invoke`, seven more waiting on a `DispatcherOperation`.

```
System.Windows.Threading.DispatcherOperation.Wait
System.Windows.Threading.Dispatcher.InvokeImpl
System.Windows.Threading.Dispatcher.Invoke(System.Action)
SysManager.ViewModels.StartupViewModel+<ScanAsync>.MoveNext()
```

`StartupViewModel.ScanAsync` shows the shape:

```csharp
if (System.Windows.Application.Current?.Dispatcher is { } dispatcher)
{
    dispatcher.Invoke(() =>
    {
        _allEntries.Clear();
        _allEntries.AddRange(sorted);
        ApplyFilter();
    });
}
else
{
    _allEntries.Clear();
    ...
}
```

The `else` exists for "no `Application`", which is the case the author had in mind. The case that hangs is the third one, and it has no branch: an `Application` exists and its dispatcher is dead.

### Where they are

Ten sites in six files, against 19 uses of `BeginInvoke` in seven others. The async form is already the house style; these are the exceptions.

| file | line |
|---|---|
| `Services/ToastService.cs` | 24 |
| `ViewModels/StartupViewModel.cs` | 99 |
| `ViewModels/ServicesViewModel.cs` | 112 |
| `ViewModels/DnsHostsViewModel.cs` | 94, 103, 454, 464, 473 |
| `ViewModels/GamingProfileViewModel.cs` | 230 |
| `ViewModels/BulkInstallerViewModel.cs` | 158 |

`ToastService.Show` is the one that matters most: 39 call sites reach it, most of them the last line of an async scan. It blocks a caller in order to raise a notification, which is the least defensible wait in the list.

```csharp
public void Show(string title, string detail, int autoHideMs = 5000)
{
[...]

PR #2155, opened 2026-09-08
fix: ten dispatcher marshals waited for a UI thread that may never run them

### LINKED ISSUE #2141: [Bug]: Integration test run aborts partway, and blame names the same test both times

laurentiu021:
Ten sites marshalled to the WPF dispatcher with the synchronous `Invoke`, each written this way:

```csharp
if (Application.Current?.Dispatcher is { } d) d.Invoke(Update); else Update();
```

That guard asks whether an `Application` **exists**, which is not the question. An `Application` can exist while nothing is pumping its dispatcher, and a synchronous `Invoke` then waits on a queue no one will drain, forever. The `else` covers "no Application" — the case the authors had in mind. The case that hangs has no branch at all.

Measured, not inferred: once #2151 removed 1,403 leaked threads of noise from the integration suite's hang dump, ten threads were parked in `Dispatcher.Invoke` and seven more waiting on a `DispatcherOperation`, under `StartupViewModel.ScanAsync` and `ServicesViewModel.RefreshAsync`. `StaHelper.Run` creates the `Application` on an STA thread and lets it exit, so from the first UI test onward `Application.Current` is set and its dispatcher belongs to a thread that no longer exists.

In the app the UI thread does pump, so these normally returned. That is what made it a latent deadlock rather than a visible bug: a blocking marshal looks harmless until something upstream waits on the work doing it.

### `Helpers/UiThread.Post`, which is not a new invention

`NetworkSharedState.InvokeOnUi` was already exactly this shape. The helper is that extracted, and `InvokeOnUi` now calls it rather than keeping a second copy that could drift. It keeps passing `DispatcherPriority.Background` explicitly, because a chart buffer should yield to input and a status line should not — that stays a per-caller decision.

**Eight sites post. Two await `Dispatcher.InvokeAsync`,** because the next statement reads what the action wrote:

- `StartupViewModel` reads `_allEntries.Count` on the following line.
- `ServicesViewModel`'s `finally` clears `IsBusy`, so posting would stop the spinner before the list it is waiting for appeared.

Awaiting keeps the ordering without parking a thread: an un-resumed continuation costs nothing, a blocked thread costs a thread.

**`ToastService.Show` keeps its no-window early return** rather than folding it into `Post`. `Post` runs inline when there is no dispatcher, which is right for a property assignment and wrong here — there is nothing to show a toast in, and running the body would construct a `DispatcherTimer` on a thread whose dispatcher never pumps. No window means no toast, which is what it did before.

**Where exceptions land does change.** With `Invoke`, an exception from the action surfaced on the calling thread where a surrounding `catch` could see it; posted, it surfaces on the dispatcher thread. Every converted action only assigns properties or raises an event, so nothing was relying on catching it — and the two whose actions run inside a meaningful `try` are precisely the two that await, so their `catch` still works.

### Red/green, seven mutations

| | | |
|---|---|---|
| baseline | | GREEN |
| M1 | ToastService back to a blocking `Invoke` | RED, names ToastService.cs |
| M2 | StartupViewModel back to a blocking `Invoke` | RED, names StartupViewModel.cs |
| M3 | the same, behind a differently-named local | RED via the alias half |
| M4 | `NamedDispatcherInvoke` matches nothing | RED via its positive control |
| M4b | it stops allowing a null-forgiving receiver | RED via its positive control |
| M5 | `AsyncMarshal` matches nothing | RED on the population floor |
| M6 | `DispatcherLocal` stops capturing the name | RED via its positive control |
| restored, sha verified, rebuilt | | GREEN |

**M3 found a real defect in the first version of the guard**, which is the whole reason to write that mutation. It injected `ui!.Invoke(…)` — the null-forgiving form — and the alias pattern allowed only `?`. It passed. Both patterns now accept `[?!]` and both carry a positive control for it, so M4b fails if that is ever narrowed again.

Note which mechanism protects what: the **population floor** protects `AsyncMarshal`, and the **positive controls** protect the two receiver patterns, because a regex that matches nothing produces no offenders and the floor cannot see it. Both are exercised above rather than assumed.

### What this does not claim

It does **not** close #2141. The last prediction of that kind, in #2151, was wrong — the leak turned out to be hiding the hang rather than causing it. What is established is that these ten calls block forever in a host with no pump, and that threads were demonstrably parked in them. Whether the never-completing test is downstream of one is what this PR's own integration run answers, and I will report that measurement here either way.

### Verification

- 110 green, 1 red over every guard in `ArchitectureTests`; the red is the throwaway local runner's own file, which is not in this repository
- `dotnet format --verify-no-changes` clean on both projects, after it caught the new file being written with LF endings
- Leak scan: 43 patterns over 17,023 bytes of added lines, 0 hits
- ARCHITECTURE.md gains `UiThread` in the utilities list, with why the old guard was the wrong question

Part of #2152.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -617,6 +617,16 @@ Key utility classes that don't fit neatly into Services or ViewModels (not an ex
   prevent because nothing holds those files open yet.
 - `RecycleBinHelper` — empties the Recycle Bin via the shell API; shared by Deep
   Cleanup and the One-Click Tune-Up so the interop has one source of truth.
+- `UiThread` — the one way background work updates the UI. Runs the action inline
+  when already on the UI thread and posts it otherwise, so the caller never waits
+  for the dispatcher. It replaced ten hand-written synchronous marshals, each
+  guarded on whether an `Application` existed — which is not the question, because
+  an `Application` can exist while nothing pumps its dispatcher and a blocking
+  `Invoke` then waits on a queue no one drains. The two callers that need the
+  update to have landed before their next statement await
+  `Dispatcher.InvokeAsync` instead: an un-resumed continuation costs nothing, a
+  blocked thread costs a thread. A fitness function asserts nothing marshals
+  synchronously, including via a short local holding the dispatcher.
 - `MarkdownTextBlock` — lightweight Markdown-to-WPF inline renderer.
 - Value converters: `EqualityConverter`, `IntGreaterThanZeroConverter`,
   `ValueConverters` (boolean/visibility/inverse helpers).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,24 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
-## [1.78.3] - 2026-09-08
+## [1.78.4] - 2026-09-08
+
+Ten places in SysManager updated the screen by handing the work to the window and then standing still until
+the window had done it. Nothing you would have noticed, because the window is normally listening — but it is
+the shape of freeze that only happens when something else has gone wrong, which is the worst time for a
+second problem. They now hand the work over and carry on.
+
+### Fixed
+- **Background work no longer waits on the window to catch up.** Ten spots did, including the notification
+  popup, which is reached at the end of most scans. A scan that has finished has nothing to gain from waiting
+  for its "done" message to appear, and if the window were ever waiting on that scan, both would have waited
+  forever. Two of the ten legitimately need the screen updated before they continue, and those now wait in a
+  way that cannot freeze anything.
+
+### Added
+- **A check that stops the pattern coming back.** It recognises the roundabout way of writing it too — the
+  form this codebase actually used, where the window is stored under a short name first — because the version
+  that only caught the obvious spelling let a test case straight through.
 
 On File Shredder and Shortcut Cleaner the line at the bottom of the page sat closer to the window edge than
 the same line does on every other tab. A few pixels, but it is the kind of unevenness you notice without
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,8 @@ second problem. They now hand the work over and carry on.
   form this codebase actually used, where the window is stored under a short name first — because the version
   that only caught the obvious spelling let a test case straight through.
 
+## [1.78.3] - 2026-09-08
+
 On File Shredder and Shortcut Cleaner the line at the bottom of the page sat closer to the window edge than
 the same line does on every other tab. A few pixels, but it is the kind of unevenness you notice without
 being able to say what is wrong.
--- a/SysManager/SysManager.Tests/ArchitectureTests.cs
+++ b/SysManager/SysManager.Tests/ArchitectureTests.cs
@@ -8506,6 +8506,129 @@ public partial class ArchitectureTests
     private static partial Regex BlockingWaitInDispose();
 
     /// <summary>
+    /// Nothing marshals to the dispatcher with the synchronous <c>Invoke</c>. Post it, or await
+    /// <c>InvokeAsync</c> when the next statement reads what the action wrote.
+    /// </summary>
+    /// <remarks>
+    /// Ten sites across six files did, each behind <c>if (Application.Current?.Dispatcher is { } d)</c> —
+    /// a guard that asks whether an <see cref="System.Windows.Application"/> EXISTS, which is not the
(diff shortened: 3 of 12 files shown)


laurentiu021:
## Measured on this PR's integration run: the hang is gone

The body promised this measurement either way. This time the prediction held.

```
Failed!  - Failed: 6, Passed: 607, Total: 613, Duration: 6 m 46 s
Data collector 'Blame' message: All tests finished running, Sequence file will not be generated.
```

| | before | after |
|---|---|---|
| outcome | `Test Run Aborted` | ran to completion |
| tests that finished | 201 of 202 recorded | **613 of 613** |
| never-completing test | `DeepCleanupViewUiTests.NavItems_DeepCleanup_AfterCleanup` | none |
| blame sequence file | generated, naming the stuck test | not generated — nothing was stuck |

That last line is the cleanest statement of it: `--blame-hang` only writes a sequence file when a test fails to complete, and it did not write one. So the ten blocking marshals were the hang, and #2141 is answered.

Three quarters of the suite had never executed. 613 total against 202 recorded before means **411 tests ran for the first time here**, which is what produced the six failures below.

## The six failures are pre-existing, and I am not fixing them in this PR

Both classes were unreachable before, so neither is a regression from this change. Establishing that mattered more than the count, so here is the evidence for each.

### Five: thread-affine app resources against a fresh STA thread per test

```
System.Windows.Markup.XamlParseException : Set property 'System.Windows.FrameworkElement.Style' threw an exception.
---- System.InvalidOperationException : The calling thread cannot access this object because a different thread owns it.
```

`AboutViewUiTests` ×4 and `DeepCleanupViewUiTests.View_Instantiates_OnStaThread`. `AppResources.Ensure()` creates the `Application` on whichever STA thread reaches it first, and the brushes and styles in its `Resources` are `DispatcherObject`s owned by that thread. `StaHelper.Run` starts a **new** thread per call, so every later view instantiation resolves app-scope resources belonging to a thread that is not only different but already gone.

Nothing in this PR touches `StaHelper`, `AppResources` or any resource. The fix is one STA thread for the suite, or frozen resources — a test-infrastructure change with its own review.

### One: a synchronous assertion on an asynchronous append

`WindowsUpdateViewModelTests.RunnerLineReceived_AppendsToConsole` raises `LineReceived` by reflection from the MTA test thread, then asserts `vm.Console.Lines.Count >= 1` on the next line. `ConsoleViewModel.Append` marshals with `BeginInvoke` and returns, so the count is still zero when the assertion runs.

That `BeginInvoke` is not mine: `ConsoleViewModel.cs` was last changed in #1873 and this branch does not touch it. The test is also order-dependent — it passes in isolation, because with no `Application` the marshal is skipped and `Append` runs inline, and fails after any UI test has created one. Hidden coupling of exactly the kind a suite that never finished could not have surfaced.

## Both filed

Filed separately rather than folded in here, so the fix that stopped the hang stays reviewable on its own and each of these gets the reproduction it needs.

Issue #2157, opened 2026-09-08
[Bug]: Tests — one console test is order-dependent, asserting synchronously on an async append


laurentiu021:
## Problem

`WindowsUpdateViewModelTests.RunnerLineReceived_AppendsToConsole` asserts synchronously on something the code appends asynchronously, so it passes or fails depending on whether an earlier test happened to create an `Application`.

The test:

```csharp
del!.DynamicInvoke(Models.PowerShellLine.Output("wu test"));

Assert.True(vm.Console.Lines.Count >= 1);
```

What it calls:

```csharp
public void Append(PowerShellLine line)
{
    if (Application.Current?.Dispatcher.CheckAccess() == false)
    {
        Application.Current.Dispatcher.BeginInvoke(() => Append(line));
        return;
    }
    …
    Lines.Add(line);
}
```

Raised by reflection from the MTA test thread, so `CheckAccess()` is false, so `Append` queues the work and returns. The count is still zero on the next line.

**It is the order dependency that makes this worth filing, more than the failure.** With no `Application` the guard is false, `Append` runs inline, and the test passes — which is what it does in isolation. Let any UI test run first and an `Application` exists for the rest of the process, so the same test fails. It has been both, in the same suite, depending on scheduling.

## Why it surfaced now

The suite aborted partway before #2155, with 202 of 613 tests recorded, and this was not among them. 411 tests ran for the first time in that PR's run. `ConsoleViewModel.Append` has used `BeginInvoke` since long before — the file was last changed in #1873 — so nothing about the behaviour is new.

## Steps to reproduce

Run the integration suite. One failure: `Assert.True() Expected: True Actual: False`.

## Expected behavior

The test asserts the append happened, and gets the same answer whichever order the suite runs in.

## Notes on a fix

The assertion has to wait for the dispatcher to run the queued work rather than reading immediately. That needs a pumped dispatcher to wait on, which is the same prerequisite as the five STA-thread failures — so it is worth fixing after that, not before, or the wait has nothing to wait for.

Two things NOT to do:

- **Do not make `Append` synchronous.** It marshals for a reason: `LineReceived` arrives from PowerShell on a pool thread and `Lines` is bound. And a blocking marshal is the defect that was just removed from ten other places.
- **Do not sleep.** A test that waits on wall-clock time is the flake this one would become.

[...]

Issue #2156, opened 2026-09-08
[Bug]: Tests — a fresh STA thread per test cannot touch the Application's own resources


laurentiu021:
## Problem

Five integration tests fail because `StaHelper.Run` starts a **new** STA thread for every call, while the `Application` those tests share was created on the first one.

```
System.Windows.Markup.XamlParseException : Set property 'System.Windows.FrameworkElement.Style' threw an exception.
---- System.InvalidOperationException : The calling thread cannot access this object because a different thread owns it.
```

`AppResources.Ensure()` creates the `Application` on whichever STA thread reaches it first. The brushes and styles in its `Resources` are `DispatcherObject`s, owned by that thread and unusable from any other. `StaHelper.Run` then hands the next test a fresh thread:

```csharp
public static void Run(Action action)
{
    var t = new Thread(() => { … });
    t.SetApartmentState(ApartmentState.STA);
    t.IsBackground = true;
    t.Start();
    t.Join();
}
```

So the second view instantiation onward resolves app-scope resources belonging to a thread that is not merely different but has already exited. The first such test passes; every later one fails at the first `Style` or `BorderBrush` it touches.

### Which tests

- `AboutViewUiTests.View_Instantiates_OnStaThread`
- `AboutViewUiTests.View_BindsToAboutViewModel`
- `AboutViewUiTests.View_BindsToFullMainVm_AboutProperty`
- `AboutViewUiTests.View_UsesCurrentVersion_FromVm`
- `DeepCleanupViewUiTests.View_Instantiates_OnStaThread`

The pattern also predicts which ones pass: `DeepCleanupViewUiTests.View_BindsToDeepCleanupViewModel` passed in the same run, because whichever test got the thread that created the `Application` is fine.

## Why it surfaced now and not before

It did not surface because the suite never got there. Before #2155 the run aborted partway with 202 tests recorded; it now completes 613, so **411 tests ran for the first time**. These five are among them. Nothing about them changed.

## Steps to reproduce

Run the integration suite. Five failures, all the same inner exception.

## Expected behavior

A test that instantiates a view resolves app-scope resources successfully, whichever order the tests run in.

## Notes on a fix

Two shapes, and the first is much better.

[...]

PR #2162, opened 2026-09-08
fix: a save was lost when anything held the file open for a fraction of a second

### LINKED ISSUE #2160: [Bug]: Cross-app — a transient file lock loses a save, and it blocked the v1.78.5 release
## Problem

`AtomicFile.Swap` gives up the moment `File.Replace` is refused, and a transient lock is enough to refuse it:

```csharp
private static void Swap(string temp, string path)
{
    if (File.Exists(path))
        File.Replace(temp, path, destinationBackupFileName: null);
    else
        MoveIntoPlace(temp, path);
}
```

No retry. `File.Replace` needs to delete the destination, and anything holding that file open — an antivirus scanner mid-read, a backup agent, a search indexer — makes it throw `IOException: Unable to remove the file to be replaced`. `AtomicFile` is the single way every service persists user data, so that is a lost save of a preset, a profile, a schedule or a history file. Callers log a failed save at Debug, so it is silent.

**Failing safely is by design and is not the complaint.** The helper exists so the previous file survives byte-for-byte rather than being truncated, and the test directly below the one that failed drives that path on purpose by holding the destination with `FileShare.None`, describing it as "what an antivirus scanner or a backup agent does to a file it is reading". The gap is narrower: a lock that would be gone in 50 ms currently costs the user their save, when the file's own documentation already names the cause.

## Evidence that it happens, not just that it could

It blocked a release. The `Release` workflow for **v1.78.5** failed at its unit-test gate:

```
SysManager.Tests.AtomicFileTests.WriteAllText_LeavesNoTemporaryFileBehind [FAIL]
  System.IO.IOException : Unable to remove the file to be replaced.
     at System.IO.FileSystem.ReplaceFile(...)
     at SysManager.Helpers.AtomicFile.Swap(...) line 232
     at SysManager.Helpers.AtomicFile.Write(...) line 129
     at SysManager.Helpers.AtomicFileTests.WriteAllText_LeavesNoTemporaryFileBehind() line 76
Failed! - Failed: 1, Passed: 5161, Total: 5162
```

The tag `v1.78.5` existed with no release behind it until the gate was re-run, which then passed with no code change. So: a Windows machine with the usual background services, two sequential writes to the same path, and the second one was refused inside a 55-second run. That is the production scenario, on the production code path, with a stack trace.

The test does nothing exotic:

```csharp
AtomicFile.WriteAllText(path, "{}");
AtomicFile.WriteAllText(path, "{\"second\":true}");
```

Something opened the file written by the first call before the second call's `Replace` reached it.

## Steps to reproduce

Non-deterministic by nature. It reproduced once in CI out of many runs. Deliberately: hold the destination open with `FileShare.None` from another thread, release it after ~100 ms, and call `WriteAllText` — the swap fails today where a retry would succeed.

## Expected behavior

A save that a brief transient lock would have blocked succeeds. A save blocked by something that is not transient still fails, still without truncating the previous file.

## Notes on a fix

A bounded retry around `Swap`, on `IOException` only, with a small number of short delays — three attempts a few tens of milliseconds apart is the usual shape for this on Windows and is enough for a scanner's read window. Then give up exactly as now.

Three things to be careful about:

- **Only `IOException`, and only around the swap.** `UnauthorizedAccessException` is a permission problem and will not improve on the second attempt; retrying it just delays the failure.
- **Do not retry the write.** The temp file is already written and flushed; re-running the caller's `writeTemp` would double the work and change what is being swapped in.
- **The delay must be injectable**, or the test for it is a sleep, and a test that waits on wall-clock time is the next flake. The project already does this elsewhere — `TimeProvider` for time-dependent code.

The pleasant property: fixing the production path fixes the test, because the test is exposed to the same condition for the same reason. Quarantining the test instead would hide the evidence and leave the user's save exposed.

Worth checking in the same pass whether `MoveIntoPlace` — the first-creation branch — has the same exposure.

## Affected area

Cross-app (`Helpers/AtomicFile`, used by every service that persists user data)

laurentiu021:
`AtomicFile.Swap` gave up the moment `File.Replace` was refused:

```csharp
private static void Swap(string temp, string path)
{
    if (File.Exists(path))
        File.Replace(temp, path, destinationBackupFileName: null);
    else
        MoveIntoPlace(temp, path);
}
```

`File.Replace` has to **delete** the destination, so anything holding that file open refuses it — an antivirus scanner mid-read, a backup agent, Windows Search. `AtomicFile` is the single way every service persists user data and callers log a failed save at Debug, so a lock that would have been gone in 50 ms cost the user their volume preset, gaming profile, maintenance schedule or history, without a word.

### Found by it happening, not by reading

It failed the `Release` workflow's unit gate for **v1.78.5**:

```
AtomicFileTests.WriteAllText_LeavesNoTemporaryFileBehind [FAIL]
  System.IO.IOException : Unable to remove the file to be replaced.
     at AtomicFile.Swap(...) line 232
```

Two sequential writes to the same path, on a CI runner with the usual background services. The tag `v1.78.5` existed with **no release behind it** until the gate was re-run, which passed with no code change. Production path, production exception, stack trace.

### Three attempts over ~155 ms, and deliberately narrow

Delays rise 15/40/100, so a slower scanner gets a longer second chance without making the common case slow.

- **`IOException` only.** An `UnauthorizedAccessException` is a permission problem and will not improve on the second attempt; retrying it only delays the error the caller needs.
- **The caller's write of the temp is not retried.** It has already succeeded and been flushed; re-running it would change what is being swapped in.
- **The exists/replace branch is re-evaluated every attempt**, not decided once. A destination deleted between attempts would otherwise leave `File.Replace` throwing `FileNotFoundException` — an `IOException` — for the rest of the budget, when the move branch would have succeeded immediately.
- **Failing safely is unchanged.** After the budget the exception propagates exactly as before and the previous file is intact byte-for-byte. The retry stops a transient refusal from being treated as permanent; it does not soften the promise.

The pause is a parameter, so the test drives the retry with **no sleeping and no timing**: its callback is what releases the lock, which makes the callback the synchronisation point. Holding the file for a wall-clock interval and hoping the retry lands inside it is the flake this avoids.

### The guard grew a wiring half, and is renamed to say so

It asserted the flush order and would not have noticed the production path losing the retry. The retry lives on a three-argument overload so it can be tested; the two-argument form is a one-line delegation, and nothing stopped that delegation being replaced by a bare `File.Replace` again — it would compile, pass every test that calls the overload directly, and quietly restore the defect. That is the #2149 shape exactly, so M2 mutates it.

### Red/green, four mutations

| | | |
|---|---|---|
| baseline | | GREEN |
| M1 | no retry at all, one attempt as it shipped | RED — the regression test, failing with the real production error: *"The process cannot access the file because it is being used by another process"* |
| M2 | production path bypasses the retrying overload | RED on the guard, the only thing that can see it |
| M3 | retry `UnauthorizedAccessException` too | RED on the guard |
| M4 | unbounded retry, no ceiling | RED on the never-clears test |
| restored, sha verified, rebuilt | | GREEN |

M4 is worth a note: it does **not** hang. `SwapRetryDelays[attempt]` overruns and it throws `IndexOutOfRangeException`, which the test catches as the wrong exception type. Different failure, still caught.

The never-clears test asserts the delays are a **rising sequence** rather than the literal 15/40/100. Restating production constants in a test makes it pass whatever they become — the mistake made once already with the health-score weights.

### Verification

- 117 green, 1 red across every `ArchitectureTests` guard plus the AtomicFile suite; the red is the throwaway local runner's own file, which is not in this repository
- `dotnet format --verify-no-changes` clean on both projects
- Leak scan: 43 patterns over 11,317 bytes of added lines, 0 hits
- ARCHITECTURE's `AtomicFile` entry records the retry and why only `IOException`

Closes #2160.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -599,6 +599,14 @@ Key utility classes that don't fit neatly into Services or ViewModels (not an ex
   `FlushOntoDevice` alone, because replacing the app's own executable must not
   inherit the outgoing build's attributes. A fitness function asserts nobody
   swaps a file into place without one of the two.
+  The swap retries a refused attempt three times over ~155 ms, because
+  `File.Replace` has to delete the destination and anything holding that file
+  open — an antivirus scanner mid-read, a backup agent, Windows Search — refuses
+  it. Without the retry a lock that would be gone in 50 ms lost the save
+  silently. Only `IOException` is retried: a permission failure will not improve
+  on a second attempt. After the budget the exception propagates exactly as
+  before and the previous file is intact, so the class's promise is unchanged.
+  The pause is a parameter so the retry can be tested without sleeping.
 - `BulkObservableCollection<T>` — `ObservableCollection` subclass that
   suppresses change notifications during bulk add/remove for UI performance
   (in `ObservableCollectionExtensions.cs`).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.78.6] - 2026-09-08
+
+If your antivirus happened to be reading one of SysManager's own settings files at the exact moment
+SysManager tried to save it, the save was abandoned — and nothing told you. Your previous settings were
+never damaged, but the change you had just made was gone. It now waits a moment and tries again.
+
+### Fixed
+- **A save is no longer lost because something else had the file open for a fraction of a second.** This
+  affects everything SysManager remembers for you: volume presets, gaming profiles, scheduled maintenance,
+  activity history, speed-test history. Antivirus scanners, backup tools and Windows Search all open files
+  briefly as they go, and a save that landed in one of those windows was simply dropped. Three quick
+  retries over about a sixth of a second cover it. A file that is genuinely locked still fails, and still
+  leaves your previous settings exactly as they were — that part has not changed.
+
 ## [1.78.5] - 2026-09-08
 
 Scheduled Maintenance offered you "TrimRam" in its dropdown and then asked you to confirm "Purge standby
--- a/SysManager/SysManager.Tests/ArchitectureTests.cs
+++ b/SysManager/SysManager.Tests/ArchitectureTests.cs
@@ -8234,7 +8234,7 @@ public partial class ArchitectureTests
     private static partial Regex DirectShutdown();
 
     [Fact]
-    public void AtomicFile_FlushesTheTempOntoTheDevice_BeforeEverySwap()
+    public void AtomicFile_FlushesBeforeEverySwap_AndTheProductionPathRetriesARefusedOne()
     {
         // Closing a handle hands the bytes to the OS write-back cache; it does not ask the drive to persist
         // them. The swap that follows is a metadata change NTFS journals, so the rename can become durable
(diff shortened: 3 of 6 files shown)


laurentiu021:
Integration run on this PR: **Failed 5, Passed 608, Total 613** — down from 6, and the 5 are exactly the STA-thread failures of #2156:

```
DeepCleanupViewUiTests.View_Instantiates_OnStaThread
AboutViewUiTests.View_Instantiates_OnStaThread
AboutViewUiTests.View_BindsToFullMainVm_AboutProperty
AboutViewUiTests.View_BindsToAboutViewModel
AboutViewUiTests.View_UsesCurrentVersion_FromVm
```

The one that dropped is `WindowsUpdateViewModelTests.RunnerLineReceived_AppendsToConsole`, and it dropped by **passing** rather than by being fixed — which is #2157's diagnosis confirmed empirically. It asserts synchronously on an async append, so it passes when it runs before any UI test has created an `Application` and fails after. Two consecutive runs of the same suite, opposite results, no code change between them touching it.

Nothing here is a regression from this change.

PR #2163, opened 2026-09-08
test: one STA thread for the suite, because the second one could not use the first's resources

### LINKED ISSUE #2156: [Bug]: Tests — a fresh STA thread per test cannot touch the Application's own resources

laurentiu021:
`StaHelper` started a fresh STA thread per call and `Join`ed it, letting the thread exit. `AppResources.Ensure()` creates the `Application` on whichever STA thread reaches it first, and the brushes and styles in its `Resources` are `DispatcherObject`s owned by that thread — so every later view instantiation resolved app-scope resources belonging to a thread that was not merely different but already gone:

```
XamlParseException : Set property 'System.Windows.FrameworkElement.Style' threw an exception.
---- InvalidOperationException : The calling thread cannot access this object because a different
     thread owns it.
```

Five failures — `AboutViewUiTests` ×4 and `DeepCleanupViewUiTests.View_Instantiates_OnStaThread`. It also explains which ones *passed*: whichever test got the thread that created the `Application` was fine. Unreachable before, because the suite aborted at test 202 of 613; visible the moment #2155 let it finish.

### Reproduced and verified locally, four ways

| | result | exit |
|---|---|---|
| thread per call, as shipped | 1 green **5 red — exactly the CI failures** | 1 |
| one shared pumped thread | 12 green | **crash 0xC0000005** |
| … + shutdown from `ProcessExit` | 12 green | **hang** (timeout) |
| … + shutdown at the end of the collection | **12 green** | **0** |

The middle two rows are the part worth reading, and neither was predictable from the issue.

**Why the crash.** A pumping dispatcher performs real layout passes over the views the tests built, and `TextBlock.MeasureOverride` reaches DirectWrite through `FontFamily`'s static initialiser. A background thread is killed where it stands when the runtime tears down, so one inside that native call as the host exits takes the process with it — *after* every test has reported, which is the most confusing possible place for a crash. The thread-per-call version could not hit it, because the thread ended before any layout pass could be scheduled.

**Why not `ProcessExit`.** That was the obvious fix and it replaced the crash with a hang: it asks WPF to run work on a thread the runtime is already tearing down. The end of the collection is the right moment — every test has reported and the runtime is still healthy — so `NetworkCollection` gained an `ICollectionFixture` that calls `StaHelper.Stop()`. Every class that calls `Run` is in that collection, verified rather than assumed, and nothing injects the fixture because a collection fixture is created and disposed whether or not a test class asks for it.

### The synchronous `Invoke` here is correct

Unlike the ten removed from the application in #2152: the contract of `Run` is "run it there and wait", the dispatcher **is** pumped by construction, and `Invoke` rethrows the action's exception on the caller's thread — which is what the hand-rolled `captured`-and-rethrow did. A test asserts that, rather than assuming the framework matches the code it replaced.

### What is traded away, deliberately

The per-call `Join` was also the isolation. On a shared pumped thread a `DispatcherTimer` started by one test keeps ticking, and work queued by an earlier test runs *before* the current action rather than being discarded. The second is stricter ordering, not looser — but a test that leaves work on the dispatcher can now be observed by the next one, which was previously impossible. That is the cost of having resources a second test can use at all.

### Tests

`StaHelperTests` pins the four properties. The regression test applies a `Style` in two separate `Run` calls, which is what threw — **applying** it is the point, because `TryFindResource` returns the object without touching a thread-affine member. That is why `AppResourcesTests.Ensure_StillResolvesStyles…` passed throughout while four view tests failed; it is renamed (it said `FromAnotherStaThread`, now untrue) with its remarks recording what it does not cover.

**What only CI can prove:** that xUnit creates and disposes the collection fixture as expected. The local runs above stand in for it by calling `Stop()` through reflection at the end. If the integration job comes back with 0 failures, that is the confirmation, and I will report it here either way.

### Verification

- 111 green, 1 red across every `ArchitectureTests` guard; the red is the throwaway local runner's own file, which is not in this repository
- Leak scan: 43 patterns over 13,599 bytes of added lines, 0 hits

`test:` — no behaviour change in the app, no version bump, no CHANGELOG, no release.

Closes #2156.

### DIFF
--- a/SysManager/SysManager.IntegrationTests/AppResourcesTests.cs
+++ b/SysManager/SysManager.IntegrationTests/AppResourcesTests.cs
@@ -37,17 +37,22 @@ public class AppResourcesTests
     }
 
     /// <summary>
-    /// A second <c>Ensure()</c>, on a second STA thread, still leaves the styles resolvable.
+    /// A second <c>Ensure()</c> still leaves the styles resolvable.
     /// </summary>
     /// <remarks>
-    /// This is the shape the suite actually runs: each view test spawns its own STA thread and calls
-    /// <c>Ensure()</c>, so all but the first take the "an Application already exists" path. The old copies
-    /// treated that as proof the resources were loaded and returned without checking, which is how a failed
-    /// load in the first caller became a <c>XamlParseException</c> in the fourth. Order-independent — it
-    /// establishes its own first call rather than relying on another test having run.
+    /// This is the shape the suite actually runs: every view test calls <c>Ensure()</c>, so all but the
+    /// first take the "an Application already exists" path. The old copies treated that as proof the
+    /// resources were loaded and returned without checking, which is how a failed load in the first caller
+    /// became a <c>XamlParseException</c> in the fourth. Order-independent — it establishes its own first
+    /// call rather than relying on another test having run.
+    /// <para>Was named <c>…FromAnotherStaThread</c>, which stopped being true when the suite moved to one
+    /// shared STA thread (#2156). It also never covered the failure that rename exposed: resolving a Style
+    /// does not touch a thread-affine member, so this passed while four view tests threw. What does cover
+    /// it is <c>StaHelperTests.Run_AStyledElementBuildsInTwoSeparateCalls</c>, which applies the style.
+    /// </para>
     /// </remarks>
     [Fact]
-    public void Ensure_StillResolvesStyles_OnASecondCallFromAnotherStaThread()
+    public void Ensure_StillResolvesStylesOnASecondCall()
     {
         StaHelper.Run(AppResources.Ensure);
 
--- a/SysManager/SysManager.IntegrationTests/StaHelper.cs
+++ b/SysManager/SysManager.IntegrationTests/StaHelper.cs
@@ -2,7 +2,7 @@
 // Author: laurentiu021 · https://github.com/laurentiu021/SystemManager
 // License: MIT
 
-using System.Windows.Threading;
+using System.Collections.Concurrent;
 
 namespace SysManager.IntegrationTests;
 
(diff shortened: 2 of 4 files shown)


laurentiu021:
## Blocked: this crashes the CI test host. Do not merge.

The integration job reports `pass` (the job is `continue-on-error`), and the log says otherwise:

```
The active test run was aborted. Reason: Test host process crashed
Passed! - Failed: 0, Passed: 329, Total: 329, Duration: 15 s
Test Run Aborted.
The test running when the crash occurred:
SysManager.IntegrationTests.StaHelperTests.Run_AStyledElementBuildsInTwoSeparateCalls
```

So on the CI runner, building a styled `Border` on the shared **pumping** thread takes the host down — during the test, not at shutdown. Locally the same test passes and the process exits 0. The difference is almost certainly the environment: a pumping dispatcher schedules real layout passes, `TextBlock.MeasureOverride` reaches DirectWrite through `FontFamily`'s static initialiser, and that is what produced `0xC0000005` locally at teardown. On a headless runner it happens immediately instead of at exit.

This is worse than what it fixes. Five failing tests became a crashed host with 284 tests never run.

### The record, four attempts

| | local | CI |
|---|---|---|
| thread per call, as shipped | 1 green **5 red** — the failures in #2156 | 613 run, 5 failed |
| one shared pumped thread | 12 green, **crash 0xC0000005 at exit** | not tried |
| … + shutdown from `ProcessExit` | 12 green, **hang** | not tried |
| … + shutdown at end of collection | **12 green, exit 0** | **host crashed after 329** |

Every attempt fixed the thing in front of it and moved the failure somewhere new, which is the signal that the approach rather than the details is wrong. **Pumping is the common factor**, and pumping is what I reached for because cross-thread `Dispatcher.Invoke` requires the target thread to pump.

### What I think the answer is, not applied

Keep one shared STA thread and **do not pump a dispatcher on it**. Give it a plain work loop — a blocking queue the thread drains — and have `Run` enqueue and wait. That fixes #2156's actual cause, which is resource *ownership by a live thread*, without buying WPF's layout machinery along with it. It is a different design, not a fourth patch: no `Dispatcher.Run`, no eager layout, no DirectWrite, and `DispatcherTimer`/`InvokeAsync` stay as inert as they are today — which is the status quo, so nothing regresses.

Marking this draft. `main` is unaffected: nothing here is merged, and the five failures remain exactly as they were.

laurentiu021:
## Rewritten: work queue instead of a pumped dispatcher

The shared STA thread was right; pumping it was not. Sharing the thread is what fixes #2156 — the resources belong to the thread that created the `Application`, and a thread that exits leaves five later view tests resolving resources nobody owns. `Dispatcher.Run()` was how I shared it, and that was the mistake.

**#2156 never needed a message loop.** It needed the owning thread to stay alive. The thread now drains a `BlockingCollection<Action>`, so nothing schedules layout, nothing reaches DirectWrite, and there is nothing in flight to interrupt at exit — no shutdown hook at all.

`DispatcherTimer` and `InvokeAsync` stay as inert here as they were with a thread per call. That is the status quo, preserved on purpose, and it is why this cannot regress anything the old version supported.

### Removed with the pumping

- The collection fixture — nothing left to shut down.
- `Run_TheThreadPumps_SoPostedWorkRuns` — it asserted a property this design deliberately does not have, and keeping it would have been the test dictating the architecture.

Replaced by `Run_RunsOnAnStaThread`, which pins the one property of the thread no other test here would notice going missing: WPF refuses to construct a visual on an MTA thread, and that failure is far less legible than the one being fixed.

### Local result

**13 green, exit 0**, including all five tests that fail on `main`. No crash, no hang.

What only CI can answer is whether the host survives the full 613 — which is exactly what the previous attempt got wrong locally, so I am not calling this done until the integration job says so, and I will report the number either way.

The scanner also needed a fix: it compared `main...HEAD` only, so on a branch that already had a commit it read the version this rewrite replaces rather than the one it introduces. It now scans both, and reports the byte count of each.

laurentiu021:
## Green. The integration suite passes in full for the first time.

```
Passed!  - Failed: 0, Passed: 618, Total: 618, Duration: 6 m 4 s
```

| | outcome |
|---|---|
| before #2155 | aborted at 202 recorded, 201 completed |
| after #2155 | 613 run, **6 failed** |
| pumped dispatcher (this branch's first attempt) | **host crashed** after 329, 284 never run |
| work queue (this branch now) | **618 run, 0 failed** |

618 rather than 613 is the five `StaHelperTests` added here.

So the answer to the open question is yes: the host survives the full run, and it survives it without a shutdown hook, because there is nothing in flight to interrupt. That was the specific thing the pumped version got wrong locally and I could not verify without CI.

Worth recording that the four-attempt sequence was not wasted motion — each attempt eliminated a hypothesis, and the third one failing *only in CI* is what identified pumping as the cause rather than shutdown timing. Locally all three pumped variants could be made to look fine.

`WindowsUpdateViewModelTests.RunnerLineReceived_AppendsToConsole` passed here too, but it is **not** fixed — #2157 stays open. It has now passed in two runs and failed in one with nothing between them touching it, which is the order-dependence that issue describes. A suite at 0 failures makes that one harder to notice, not less real.

PR #2165, opened 2026-09-08
fix: an elevated tab started a new PowerShell helper for every single call

### LINKED ISSUE #2149: [Bug]: Elevated PowerShell leaks a runspace per call, and opening one can hang forever

laurentiu021:
Every elevated `RunAsync` built a runspace, and elevated means an **out-of-process** one: a `powershell.exe` 5.1 child, spawned, plus a remoting handshake with it, then thrown away. That is the slow part of the call — and the services that use it call in **bursts**: `DnsService` six times, `EdgeOneDriveService` four, TaskScheduler / RestorePoints / ScheduledMaintenance three each. Six spawns where one would do.

An elevated runner now keeps its runspace and reuses it. **Three properties make that safe, and each has a test that fails without it.**

| | why it is not optional |
|---|---|
| **State re-checked on every lease** | A cached runspace can be broken from outside — the child killed by the user or by cleanup, the channel dropped — and one that is not `Opened` cannot run a pipeline at all. Without the re-check, one broken child makes every later call on that consumer fail for the rest of the session: far worse than the cost being removed. |
| **Pipelines serialised behind a gate** | A runspace runs one at a time, so where a runspace per call made concurrent calls independent, a shared one makes them a conflict. Cheap in practice — Transient registration means one runner per consumer, and the consumers that make several calls already serialise them — so the gate is the guarantee rather than a dependency on that staying true. |
| **Evicted after ~20 s idle** | The half that keeps this a latency win rather than a memory trade. See below. |

### "One runspace per session" is the obvious reading and it is the wrong one

Nothing disposes most of these runners — **nine are constructed directly in `MainWindowViewModel`'s designer graph** and live as long as the window. A session-long cache would therefore mean a dozen `powershell.exe` processes resident for the whole run, tens of MB each, in an app whose entire pitch is making a PC feel faster. The benefit is in the burst; eviction takes the cost back once you have moved on.

`Dispose` releases it deterministically where a consumer *is* disposed. It is not the only release path and must not be, for the nine-runners reason.

**Scoped to the elevated path, asserted rather than described.** The in-process runspace starts no child and opens in a fraction of the time, so caching it would add a lifetime to reason about for almost no gain. `UnelevatedRunner_StillBuildsARunspacePerCall` is what would notice that scope being widened.

**The isolation boundary is unchanged, and reuse cannot weaken it:** the trusted machine-only `PSModulePath` is applied to the child's environment when it is spawned, so a cached child keeps the value it was given. `RunAsync_WhenElevated_IsolatesModuleDiscoveryInChildProcess` forces elevation and makes two calls, so it now exercises the reused runspace and still asserts the injected module is unreachable.

### Red/green, six mutations

| | | |
|---|---|---|
| baseline | | 8 green |
| M1 | no cache, always build fresh | RED — reuse test |
| M2 | invert the elevation check | RED — scope test |
| M3 | never arm idle eviction | RED — eviction test |
| M4 | `Dispose` leaves the cached runspace alone | RED — child-release test |
| M5 | reuse without re-checking runspace state | RED — rebuild test |
| M6 | drop the pipeline gate entirely | RED — overlap test |
| restored, sha verified, rebuilt | | 8 green |

**Three of those found real flaws, which is the reason to run them:**

1. **The concurrency test passed with the gate deleted.** Two calls started back to back do not conflict — the first has not cached anything by the time the second leases, so each builds its own runspace and nothing is shared. It needed a warm-up call first to make both callers reach for the same runspace. Without that fix the gate would have *looked* tested and been untested.
2. **`if (false)` / `if (true)` as mutations don't compile here** — they make the following statement unreachable and CS0162 is an error, so the proof reported BROKEN, which says nothing about the guard. Inverting a condition, or comparing against a state that cannot occur, mutates behaviour with a tree that still builds.
3. **Removing only the gate's `Wait`** left its `Release` over-releasing the semaphore, so every test died on `SemaphoreFullException` — red for the wrong reason. The mutation now removes both halves.

The idle-eviction test's first version polled *faster* than the idle period, and every call re-arms the countdown, so it kept the runspace alive forever and timed out. It now waits longer than the lifetime and doubles if a slow machine needs more.

**Renamed:** `RunspaceTeardown_StopsTheChildTheRunspaceWasBuiltWith` → `DisposingTheRunner_StopsTheChildItsRunspaceWasBuiltWith`, because the end of a run is no longer when the child is released. It also now asserts the child is **still alive** after the run, which is what says reuse is actually taking effect rather than merely not crashing.

### Verification

- Four projects build, 0 errors 0 warnings, on a full `--no-incremental` rebuild
- 111 green, 1 red across every `ArchitectureTests` guard; the red is the throwaway local runner's own file, which is not in this repository
- Leak scan: 43 patterns over 20,636 bytes of added lines, 0 hits
- ARCHITECTURE's `PowerShellRunner` entry records the reuse and all three safety properties

Deferred to the workstation that runs the application: how much quicker DNS & Hosts and Edge/OneDrive Remover actually feel when elevated. The mechanism is proven by count through the `createRunspace` seam rather than by timing, so this is a look rather than a question.

Closes #2149.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -184,6 +184,20 @@ Key services:
   use an in-process runspace and retain per-user modules. Elevated sessions use
   an isolated Windows PowerShell 5.1 child with module discovery limited to canonical
   machine-owned roots, avoiding process-global environment mutation.
+  An elevated runner REUSES its runspace across calls, because starting that child
+  and completing a remoting handshake with it is the slow part and the services
+  that use it call in bursts (`DnsService` six times, `EdgeOneDriveService` four).
+  Three properties make reuse safe: the runspace state is re-checked on every
+  lease, so a child killed from outside is discarded and rebuilt rather than
+  failing every later call; pipelines are serialised behind a gate, because a
+  runspace runs one at a time and a shared one turns concurrent calls into a
+  conflict; and the runspace is evicted after ~20 s idle, which matters because
+  nine runners are constructed directly in `MainWindowViewModel`'s designer graph
+  and nothing ever disposes them — without eviction a dozen `powershell.exe`
+  processes would be resident for the whole run. `Dispose` releases it
+  deterministically where a consumer is disposed. The unelevated path is
+  unchanged and still builds per call: it starts no child, so caching would add a
+  lifetime to reason about for almost no gain.
 - `WingetService` — shells out to `winget` and parses its table output.
 - `WindowsUpdateService` — drives Windows Update through the WUA COM API
   (scan, select, install) with progress reporting; backs `WindowsUpdateViewModel`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.78.7] - 2026-09-08
+
+Running SysManager as administrator made several tabs feel slower than they needed to. Every single thing
+they asked Windows to do started a fresh hidden helper first and waited for it to be ready — DNS &
+Hosts does that six times, Edge/OneDrive Remover four. Those tabs now start one helper and use it for the
+whole batch, and let it go once you have moved on.
+
+### Changed
+- **Administrator tasks reuse their helper instead of starting a new one every time.** The wait to start
+  one is the slow part, so a tab doing several things in a row is now noticeably quicker on the second
+  action onward. It is let go about twenty seconds after you stop, so nothing is left sitting in memory
+  while you use the rest of the app — and if it ever stops responding, the next action quietly starts a
+  fresh one rather than failing.
+
 ## [1.78.6] - 2026-09-08
 
 If your antivirus happened to be reading one of SysManager's own settings files at the exact moment
--- a/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
+++ b/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
@@ -12,9 +12,19 @@ namespace SysManager.IntegrationTests;
 public class PowerShellRunnerTests
 {
     /// <summary>
-    /// A run's teardown STOPS the child process the runspace was built with, rather than only dropping its
-    /// handle.
+    /// Disposing the runner STOPS the child process its runspace was built with, rather than only dropping
+    /// the handle.
     /// </summary>
+    /// <remarks>
+    /// <para><b>Was "a run's teardown", and that changed.</b> An elevated runner now REUSES its runspace
+    /// across calls rather than building one per call (#2149), so the end of a run is no longer when the
+    /// child is released — the release points are eviction after
+    /// <see cref="PowerShellRunner.IdleRunspaceLifetime"/> and disposal of the runner. This asserts the
+    /// deterministic one.</para>
+    /// <para>Elevation is forced rather than inherited so the test exercises the reuse path on any machine.
+    /// The runspace itself is in-process, so nothing here needs administrator rights; the injected child
+    /// stands in for the <c>powershell.exe</c> that the real elevated path would have started.</para>
+    /// </remarks>
     /// <remarks>
     /// #2149(a), end to end through <c>RunAsync</c>. Disposing a <see cref="System.Diagnostics.Process"/>
     /// releases the HANDLE and does not stop the process. When <c>powershell.exe</c> outlived its runspace, its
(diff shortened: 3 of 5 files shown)


laurentiu021:
Integration run on this PR: **Failed 0, Passed 623, Total 623**, including all five new tests here — and it ran with the runner **actually elevated**, which is what the CI runner is and this machine is not. So the reuse path, the state re-check, the gate, eviction and disposal were all exercised against a real out-of-process runspace with a real `powershell.exe` child, not just the injected stand-ins the local proof used.

618 → 623 is the five added tests.

Issue #2195, opened 2026-09-09
[Bug]: CI — the integration job reported "Failed!" on a run where all 625 tests passed; the hang-dump extension timed out in its own shutdown


laurentiu021:
## Problem

On the CI run for #2192 the integration job finished with `conclusion: success` at the job level while its
own summary said `Test run summary: Failed!` with `error: 1` — and every one of the 625 tests passed.

The exception is inside the hang-dump extension, after the tests were over:

```
Unhandled exception. System.TimeoutException: The operation has timed out.
   at Microsoft.Testing.Platform.Helpers.TaskExtensions.TimeoutAfterAsync(Task task, TimeSpan timeout, …)
   at Microsoft.Testing.Extensions.Diagnostics.HangDumpProcessLifetimeHandler.DisposeAsync()
```

```
Test run summary: Failed!
  error: 1
  total: 625
  failed: 0
  succeeded: 625
  skipped: 0
  duration: 24m 12s 355ms
```

A full dump was written (`SysManager.IntegrationTests_5792_hang.dmp`), so `--hangdump` fired, but no test was
named — `longRunningTestSeconds` produced no diagnostic, which it would have if a test had stalled. No test
failed. The frames are entirely in `HangDumpProcessLifetimeHandler.DisposeAsync`, i.e. the extension's own
shutdown handshake with a test host that had already finished its work.

Two things then went wrong on top of it:

- **The job still reads as green.** `gh pr checks` showed `Integration tests (non-blocking) — pass`, because
  `continue-on-error: true` forces the step's *conclusion* to success. The workflow already knows this trap
  and guards it by keying the annotation on `outcome` — and that annotation DID fire. So the warning is in
  the log and the check row says pass. Anyone reading the row alone sees a clean run.
- **It cost 24m12s against a 30-minute ceiling.** The immediately preceding run of the same job, on #2191,
  took 7m52s end to end. Another minute of drift and this becomes a `cancelled` job, which renders as `fail`
  — the exact permanently-red state the job's own comment says makes a non-blocking check get ignored.

Not caused by the PR it appeared on: #2192 changed `StartupView.xaml`, `ArchitectureTests.cs` (unit project),
and four docs. Only `AboutView` and `DeepCleanupView` have integration UI tests and nothing enumerates
`Views/`, so `StartupView.xaml` is never loaded by that suite.

## Proposed solution

Decide between two, and the choice needs measuring first — how often does this signature recur? One
occurrence is not yet a pattern.

[...]

Issue #2206, opened 2026-09-09
[Bug]: CI — the integration cancellation test ran a 30s script to completion once, and the obvious cause is refuted


laurentiu021:
## Problem

`PowerShellRunnerTests.RunAsync_SupportsCancellation` failed on the #2203 run:

```
failed SysManager.IntegrationTests.PowerShellRunnerTests.RunAsync_SupportsCancellation (31s 122ms)
  Cancellation took too long: 00:00:31.1204946
```

The elapsed time is the interesting part. The test cancels at 300 ms and the script is
`Start-Sleep -Seconds 30`, so **31.12 s means the sleep ran to completion** — cancellation had no effect at
all, rather than being slow. On a tab with a Cancel button that is a real user-facing failure, not a timing
wobble, which is why this is filed as a bug rather than as a flaky test.

The job reported `pass` at job level (`continue-on-error`), with the real summary
`Failed! total: 625, failed: 1` in the log. Same masking as #2195.

## What I ruled out — the obvious cause is NOT it

`RunOnLeasedRunspaceAsync` registers the cancellation callback **before** starting the pipeline:

```csharp
using var reg = cancellationToken.Register(() => { try { ps.Stop(); } catch (InvalidOperationException) { } });
var task = Task.Factory.FromAsync(ps.BeginInvoke<PSObject, PSObject>(null, output), ar => ps.EndInvoke(ar));
```

`LeaseRunspaceAsync()` above it takes no token and opens a runspace, which its own comment says "can take
several hundred milliseconds". With a 300 ms cancellation that token is plausibly already cancelled by the
time `Register` runs — and `CancellationToken.Register` invokes the callback synchronously when the token is
already cancelled. So `ps.Stop()` appeared to land on a `NotStarted` instance, `BeginInvoke` would then start
an unstoppable script, and the elapsed time would be the full sleep. It fits the number exactly.

**It is wrong.** I built a harness (temporary console app referencing the app project, so no `dotnet test`)
and mutated the runner three ways, rebuilding and re-measuring each time:

| Mutation | Result |
|---|---|
| `Register` before `BeginInvoke` (the shipped order) | **GREEN** — cancelled in 0.32 s, `OperationCanceledException` |
| the post-await `ThrowIfCancellationRequested` removed | **GREEN** |
| a pre-start `ThrowIfCancellationRequested` removed | **GREEN** |

[...]

Issue #332, opened 2026-05-13
[Enhancement]: Customization — Per-App Volume Mixer (EarTrumpet-style)


laurentiu021:
## Problem
Windows built-in volume mixer is basic and hard to access. Users want per-application volume control with output device routing.

## Proposed solution
- Modern per-application volume control
- Per-app output device routing (game audio to headset, music to speakers)
- Save volume presets per profile
- Visual sliders with app icons
- Quick mute per-app
- System tray integration for quick access

## Affected tab
Customization (new group)

laurentiu021:
Reopening: this feature is still planned and tracked. It is surfaced in-app as a work-in-progress page and was not yet implemented, so the closed-as-completed state was inaccurate.

laurentiu021:
Reopening — the first slice of this shipped as a **Preview** in #1298, but the full request isn't complete yet.

**Shipped in the preview (Volume Control tab, Customization group):**
- Modern per-application volume control with visual sliders and app icons
- Quick mute per-app
- Live per-app level meter
- (default playback device)

**Still to do (tracked by this issue):**
- Per-app **output-device routing** (game → headset, music → speakers) — this needs an undocumented, version-fragile Windows internal API with a silent-mute failure mode on an invalid device, so it's deliberately held back until it can be done safely
- Save **volume presets** per profile
- **System tray** quick-access flyout
- Multi-output-device coverage (beyond the default endpoint)

The tab is marked **Preview** in the sidebar while it settles in. Leaving this issue open to track the remaining scope.

Issue #331, opened 2026-05-13
[Enhancement]: Customization — Environment Variable Editor (PATH management GUI)


laurentiu021:
## Problem
Editing PATH and environment variables in Windows is painful: a single long textbox, no validation, easy to break things.

## Proposed solution
- GUI for system and user PATH + all environment variables
- Drag-reorder entries
- Detect duplicate entries (highlight)
- Detect non-existent paths (red highlight)
- Validate before apply
- Backup/restore env vars
- Search/filter

## Affected tab
Customization (new group)

laurentiu021:
Shipped in **v1.22.0**.

The Customization group's **Environment Variables** tab is now fully implemented — a proper GUI for editing User and System variables, with the dedicated PATH editor this issue asked for.

**What it does**
- Lists **User and System** variables in one grid; filter by scope and search by name or value.
- Edit values in place, and add or remove variables.
- **Dedicated PATH editor** for PATH-like variables: reorder directories (up/down), remove entries, and **strip duplicates** in one click.
- **Missing folders are highlighted** so dead PATH entries are obvious at a glance.
- **Validation before apply**: variable names are checked at the trust boundary (no `=`, no whitespace, no leading digit, no shell metacharacters).
- **Backup/restore**: a one-time JSON snapshot of every variable is written before the first change, so the original environment is recoverable.

**How it's built**
- `EnvironmentVariableService` reads/writes both scopes via `Environment.SetEnvironmentVariable`, which broadcasts `WM_SETTINGCHANGE` so new terminals pick up changes without a reboot. It mirrors the existing `HostsFileService` backup contract and `PrivacyService`'s elevation-aware writes (a denied System-scope write returns gracefully instead of throwing). PATH split/join/dedupe are pure, unit-tested helpers.
- `EnvironmentVariablesViewModel` follows the existing staged-edit pattern (baseline + pending-change count, Apply/Discard, `DialogService` confirmation, admin elevation banner). System-scope edits require administrator; User-scope edits don't.
- `EnvironmentVariablesView` reuses the standard tab layout, styles, and accessibility conventions.

**Verification**
- 20 unit tests: name validation (accept/reject, including injection-ish and overlong inputs), PATH split/join round-trips, deduplication (case-insensitive, trailing-slash aware), and backup create/no-overwrite/corrupt/missing handling.

[...]

PR #2236, opened 2026-09-10
test: stop the ping churn orphaning thousands of pings into the next test

laurentiu021:
Root cause of #2195. The annotation half of that issue (proposal 1) shipped earlier; this is the other half, and it turned out not to be the knob the issue expected.

### What was actually happening

The recurrence measurement the issue asked for first: **8 of 60 completed integration jobs**, clustering at ~26 minutes against a median of 8.2, with **3.5 minutes of headroom** before the 30-minute ceiling turns it into a `cancelled` job that renders red. So it is a pattern, not a one-off.

Then per-test durations out of the TRX of one slow run and one healthy run:

```
slow run: 625 results, 546.5 s total
fast run: 625 results, 491.8 s total
per-test slow/fast ratio: median 1.00x, p90 1.51x
```

The tests are not slower. Nine minutes of tests in both cases, against a run duration of `8m 17s` on the healthy run and `24m 15s` on the slow one — so ~15 minutes is spent outside the tests entirely, writing a 704 MB dump and then timing out in `HangDumpProcessLifetimeHandler.DisposeAsync`.

What the same comparison did show is one test carrying 58% of the suite, on **every** run:

| test | healthy run | slow run | intended `Task.Delay` |
|---|---|---|---|
| `ToggleIsEnabled_MidFlight_IsRespected` | 283.50 s | 336.74 s | **1.2 s** |

283 s is under `--hangdump-timeout 5m`. 336 s is over it. That is the whole 13%.

### Why that test costs 283 s when it sleeps for 1.2

It doesn't. It pays for the test that runs immediately before it, in both runs: `ParallelAddRemoveWhileRunning_IsThreadSafe`.

That test churns the target map for two seconds in an unthrottled loop, adding hosts drawn from `192.0.2.2-253`. `PumpAsync` fires one ping per **enabled** target per tick, so with a full map at `Interval = 120 ms` it starts on the order of 4000 real ICMP operations to unroutable addresses — every one of them abandoned mid-flight when `Stop()` cancels the token, disposing its `Ping` while the native send is outstanding.

`SysManager.IntegrationTests/xunit.runner.json` sets `maxParallelThreads: 1` and `parallelMode: none`, so all of those continuations drain through a single worker, and they drain *after* the test that started them has already returned. The arithmetic holds across the neighbours too: `ManyTargets_NoExceptions` orphans ~150 pings and costs 5.8 s (~39 ms each); the churn orphans ~4000 and the next test costs 283 s (~71 ms each).

The reason it needed a TRX diff to find is that the mechanism defeats every signal pointed at it. The dump named no test, `longRunningTestSeconds: 120` produced no diagnostic, thread-pool starvation is ruled out because the test *after* the slow one takes 5 s, and the assertion passes cleanly the whole time.

### The change

```csharp
svc.AddOrUpdate(new PingTarget("x", h, "#111") { IsEnabled = false });
```

`PumpAsync` snapshots `Targets.Values.Where(t => t.IsEnabled && …)` every tick, so a disabled target is **still enumerated** — the concurrent `AddOrUpdate`/`Remove` still races that snapshot, which is the entire reason the test exists — it is simply never pinged. `"base"` stays enabled, so the pump is genuinely pinging while the map churns underneath it, and all four assertions are untouched.

Nothing is lost in coverage: mid-pump target appearance and disappearance is covered by `PingMonitorServiceTests.AddingTargetMidPump_StartsEmitting` and `RemovingTargetMidPump_StopsEmitting`, and `PingOnceAsync`'s stale-sample guard by `DisabledTarget_EmitsNoSamples`.

`--hangdump-timeout` is deliberately **not** touched. The issue offered raising it as the safer knob, but that was on the assumption the suite had a legitimately slow test. It doesn't — after this, the slowest is ~70 s, so 5 minutes keeps its headroom and keeps meaning "something is genuinely stuck". Raising it would have hidden this instead of fixing it.

### Guard

`ThePingChurn_BuildsDisabledTargets_SoItDoesNotOrphanThousandsOfPings`, in the **blocking** unit suite rather than the integration suite it guards — a job whose symptom is that job timing out cannot be trusted to report it. It reads the churn loop out of the test's source, comment-stripped, and fails if any target built inside it is left enabled.

Red before, green after, all three failure paths, restore verified byte-for-byte:

```
baseline (unmutated): GREEN
mutation 1 churned target left enabled -> RED   "1 of the 1 PingTargets built inside the churn loop are left enabled…"
mutation 2 slice marker moved          -> RED   "the churn loop … was not found, so this guard read nothing"
mutation 3 far end of the loop removed -> RED   "the churn slice does not reach the loop's Remove call…"
3 of 3 mutations failed for the expected reason
```

Mutation 2 and 3 exist because the first version of the guard was red against correct code: an object initializer ends in `});`, so the slice terminated inside the very construction it was trying to read, and the `>= 1` vacuity floor still passed because one truncated `new PingTarget(` remained inside it. The terminator is now anchored to the start of a line, and the slice must reach the loop's `Remove` call — a truncated slice can only under-report.

### Verification

- `dotnet build` — integration and unit projects, **0 errors, 0 warnings**
- Unit suite: **5442 passed**, 0 failed (5441 + the new guard)
- `dotnet format --verify-no-changes` — clean on both changed projects
- Leak scan of the staged diff against all 43 current patterns — 0 hits over 94 added lines

The falsifiable prediction for this PR's own integration run: the suite should drop from ~8m17s to roughly 3m30s, `ToggleIsEnabled_MidFlight_IsRespected` from 283 s to about 1.5 s, and `ManyStartStopCycles_NoLeak` from 33–49 s to a few seconds. If it doesn't, the diagnosis is wrong and this should not merge.

`test:` — no release.

### DIFF
--- a/SysManager/SysManager.IntegrationTests/PingMonitorStressTests.cs
+++ b/SysManager/SysManager.IntegrationTests/PingMonitorStressTests.cs
@@ -50,33 +50,55 @@ public class PingMonitorStressTests
         svc.Start();
         Assert.True(svc.IsRunning);
 
-        using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(2));
         // The churn deliberately never touches "base" — only 192.0.2.2-253 hosts — so
-        // it survives as a fixed point we can assert on after the storm settles.
+        // it survives as a fixed point we can assert on after the storm settles. "base"
+        // stays enabled, so the pump is genuinely pinging while the map churns under it.
         //
-        // The churned targets are DISABLED, and that is load-bearing rather than tidiness.
-        // PumpAsync snapshots `Targets.Values.Where(t => t.IsEnabled …)` every tick, so a
-        // disabled target is still enumerated — the concurrent add/remove still races the
-        // snapshot, which is the entire point of this test — but it is never pinged. With
-        // them enabled, an unthrottled two-second loop filled the map with ~250 hosts and
-        // the pump then fired ~4000 fire-and-forget ICMP operations at unroutable
-        // addresses, every one of them abandoned mid-flight when Stop() cancelled the
-        // token. The suite runs maxParallelThreads=1, so those continuations all drained
-        // through a single worker AFTER this test returned, and the bill landed on
-        // whichever test ran next: ToggleIsEnabled_MidFlight_IsRespected measured 283s on
-        // a healthy run and 336s on a slow one, against 1.2s of intended Task.Delay. Above
-        // 300s it crossed --hangdump-timeout 5m, so the runner wrote a 704 MB dump and the
-        // hang-dump extension's DisposeAsync then timed out, reporting `Failed!` with
-        // `failed: 0` — 13% of runs (8 of 60 measured). "base" stays enabled so the pump
-        // is genuinely pinging while the map churns underneath it. See #2195.
-        var churn = Task.Run(() =>
+        // Two things here are load-bearing rather than tidiness, and both were measured
+        // on #2195. This test used to spin unthrottled for two seconds against a
+        // CancellationTokenSource, with the churned targets left ENABLED.
+        //
+        // ENABLED was the expensive half. PumpAsync snapshots
+        // `Targets.Values.Where(t => t.IsEnabled …)` every tick, so a disabled target is
+        // still enumerated — the concurrent add/remove still races that snapshot, which
+        // is the entire point of this test — but it is never pinged. With them enabled the
+        // map filled with ~250 hosts and the pump fired thousands of fire-and-forget ICMP
+        // operations at unroutable addresses, all abandoned mid-flight when Stop()
+        // cancelled the token. The suite runs maxParallelThreads=1, so those continuations
+        // drained through one worker AFTER this test returned, and the bill landed on
+        // whichever test ran next: ToggleIsEnabled_MidFlight_IsRespected measured 283s
+        // against 1.2s of intended Task.Delay, and 336s on a slow run. Past 300s it
+        // crossed --hangdump-timeout 5m, the runner wrote a 704 MB dump, and the hang-dump
+        // extension's DisposeAsync then timed out and reported `Failed!` with `failed: 0`
+        // on 8 of 60 runs. Disabling them moved that test to 7.64s.
+        //
+        // UNTHROTTLED was the unstable half, and disabling the targets alone did not fix
+        // it — it relocated it. The two-second spin measured ~27 MILLION add/remove
+        // operations, i.e. ~14 million ConcurrentDictionary writes a second alongside a
+        // pump snapshotting the same dictionary every tick, and the cost of that was never
+        // stable: the same runner image measured this pair at 4s + 283s, then at 217s + 8s.
+        // Interleaving is what proves thread safety, not iteration count, so the churn is
+        // bounded on OPERATIONS and yields between batches — a fixed ~20k operations
+        // spread across several pump ticks, at a cost that does not depend on how fast the
+        // machine happens to be.
+        //
+        // Yes, that is an awaited delay in a test, which ArchitectureTests.NoTestWaitsBySleeping
+        // forbids — and that guard is scoped to the blocking unit suite, deliberately, not to
+        // this project. The rule exists because a unit test must not assert how fast the
+        // machine is. Here the subject under test IS a timer: PumpAsync ticks on Interval, so
+        // the churn has to span ticks to race the snapshot at all, and Task.Yield() would
+        // finish all 20k operations before the pump ticked once. The delay is what makes this
+        // test's cost machine-INdependent rather than dependent on it.
+        const int ChurnOperations = 20_000;
+        var churn = Task.Run(async () =>
         {
             var rnd = new Random(42);
-            while (!cts.IsCancellationRequested)
+            for (int i = 0; i < ChurnOperations; i++)
             {
                 var h = $"192.0.2.{rnd.Next(2, 254)}";
                 svc.AddOrUpdate(new PingTarget("x", h, "#111") { IsEnabled = false });
                 if (rnd.Next(2) == 0) svc.Remove(h);
+                if (i % 1_000 == 0) await Task.Delay(25);
             }
         });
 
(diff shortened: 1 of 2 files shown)


laurentiu021:
The prediction in the description failed, so here is what CI actually measured and what the second commit does about it.

### The prediction, scored

I said the suite should drop from ~8m17s to ~3m30s and `ToggleIsEnabled_MidFlight_IsRespected` from 283 s to about 1.5 s, and that if it didn't, the diagnosis was wrong and this should not merge. Half of it held:

| test | before | after |
|---|---|---|
| `ToggleIsEnabled_MidFlight_IsRespected` | 283.50 s | **7.64 s** |
| `ParallelAddRemoveWhileRunning_IsThreadSafe` | 4.37 s | **217.46 s** |
| `ManyStartStopCycles_NoLeak` | 49.34 s | 11.04 s |
| suite total | 8m 17s | **7m 18s** |

So the orphaned-ping mechanism was real and removing it did move 275 s out of that test. But the cost did not disappear, it **relocated** into the test that creates it — and 217 s straddles `--hangdump-timeout 5m` exactly like the test it replaced. Net saving 59 s. That is the same defect with a new name, and it is not mergeable.

Worth recording that the execution order is byte-identical between the two runs (`#329 ParallelAddRemove → #330 ToggleIsEnabled → #331 ManyTargets → #332 ManyStartStopCycles`), so this is not a reordering artefact. `PowerShellRunnerTests.RunAsync_SupportsCancellation` also went 0.35 s → 30.67 s, which says the drain is still landing on unrelated neighbours.

### What I could and could not reproduce

A console harness that drives the exact churn loop against a real `PingMonitorService`, with the service never started so nothing touches the network:

```
shape                       elapsed     iterations    overrun
no pump, enabled #1           2.00s     27,825,219      0.00s
no pump, disabled #1          2.01s     27,174,882      0.01s
PUMP, enabled #1              2.00s     23,422,264      0.00s
PUMP, disabled #1             2.01s     27,349,587      0.01s
```

Three rounds of each, all 2.00–2.01 s. So the loop itself is not the cost and the `IsEnabled` flag does not change it — the 217 s needs real ICMP on a 4-core runner, which is not reproducible on a 32-core developer machine. I stopped there rather than guess a fourth mechanism.

What does not need reproducing is the **~27 million add/remove operations in two seconds**. That is 14 million `ConcurrentDictionary` writes a second against a pump snapshotting the same dictionary every tick, and a test with that shape has no stable cost — measured at 4 s and at 217 s on the same runner image. That is indefensible in a suite independently of which mechanism converts it into minutes, and it is what the second commit fixes.

### Second commit

The churn is bounded on **operations** instead of wall time: a fixed 20 000, spread across several pump ticks with a delay between batches. Every assertion is untouched, and `IsEnabled = false` stays, since it is what moved 275 s out of the neighbouring test.

On the delay: `ArchitectureTests.NoTestWaitsBySleeping` forbids an awaited delay in a test, and is scoped to the blocking unit suite rather than this project. That rule exists because a unit test must not assert how fast the machine is. Here the subject under test **is** a timer — `PumpAsync` ticks on `Interval`, so the churn has to span ticks to race the snapshot at all, and `Task.Yield()` would finish all 20 000 operations before the pump ticked once. The delay is what makes this test's cost machine-*in*dependent. The reasoning is at the call site so the next reader does not have to re-derive it.

The guard follows the code: the slice now anchors on the bounded `for`, and it additionally requires the throttle, because that is one deletable line. Its own needle is split the way `NoTestWaitsBySleeping` splits its needles — written whole, the literal made *that* guard report `ArchitectureTests.cs` as a test that sleeps, which is a fair thing for it to have caught.

### Verification

- 4 mutations, each RED for its own stated reason — target left enabled, slice marker moved, far end of the slice removed, throttle removed — restore byte-for-byte
- Unit suite **5442 passed**, 0 failed
- Both projects 0 errors, 0 warnings; `dotnet format --verify-no-changes` clean on both
- Leak scan of the diff against all 43 patterns — 0 hits

No prediction this time, because the last one was wrong in a way I could not have caught locally. The number to read on this run is `ParallelAddRemoveWhileRunning_IsThreadSafe`: if it is not seconds, the bound is not the answer either and this needs a different approach rather than a third patch.

laurentiu021:
The bound is the answer. Measured on this run, all four jobs green, integration included.

| test | baseline | disable targets only | + bound the churn |
|---|---|---|---|
| `ParallelAddRemoveWhileRunning_IsThreadSafe` | 4.37 s | 217.46 s | **10.01 s** |
| `ToggleIsEnabled_MidFlight_IsRespected` | 283.50 s | 7.64 s | **3.42 s** |
| `ManyStartStopCycles_NoLeak` | 49.34 s | 11.04 s | **10.04 s** |
| `ManyTargets_NoExceptions` | 4.94 s | 3.60 s | **2.22 s** |
| tests, total | 491.8 s | 432.7 s | **150.6 s** |
| run step, wall clock | 8m 17s | 7m 18s | **2m 34s** |

The number I said would falsify the diagnosis was `ParallelAddRemove`: it is 10 s, so the bound is the answer and not a third patch. Both halves were needed — disabling the targets alone relocated the cost, and bounding the loop alone would have left thousands of orphaned pings to redistribute.

The threshold this issue is about now has an order of magnitude of headroom: the slowest test in the suite is 29.6 s against `--hangdump-timeout 5m`, where before it was 283–336 s straddling it. So `--hangdump-timeout` stays at 5 minutes and keeps meaning "something is genuinely stuck", which is what the step's own comment argues for.

### An unexpected result worth a separate issue

`PowerShellRunnerTests.RunAsync_SupportsCancellation` went **0.35 s → 30.67 s → 0.33 s**.

That is #2206 — the test that once reported `Cancellation took too long: 00:00:31.12` and whose obvious cause that issue refuted with a harness. It read 30.67 s in the run where the drain redistributed onto the PowerShell tests, and 0.33 s here. The harness in #2206 always cancelled in ~0.35 s precisely because an isolated harness has no saturated worker to queue behind. I'll write that up on #2206 with these three readings rather than bury it here; the short version is that its stopwatch was measuring scheduler starvation, not a cancellation that failed.

`ElevatedPowerShell_WhenStartupWouldRestore` is the one row moving the wrong way (1.76 → 10.96 → 29.57 s) and is now the slowest test in the suite. Not touching it here — it is 29.6 s against a 300 s threshold, elevation and runspace startup are inherently variable, and I have no evidence yet about which direction that variance runs. Noting it so the next person reading these numbers does not think it went unseen.

Merging.

Issue #1577, opened 2026-08-03
[Enhancement]: Cleanup / Deep Cleanup - Deep Cleanup misses the three biggest Windows space hogs: WinSxS component store, kernel memory dumps,...


laurentiu021:
## Problem
`DeepCleanupService.BuildDefinitions()` ships 19 buckets (`Services/DeepCleanupService.cs:55-166`, verified by count). Three high-yield buckets every rival offers are absent, all verified by grep across the whole project: (1) **Kernel memory dumps** - `grep -rniE "memory\.dmp|minidump|livekernel" --include=*.cs` returns nothing. The existing "Crash dumps & error reports" bucket (lines 97-105) covers only `%LocalAppData%\CrashDumps` and the WER report queues - it does NOT cover `C:\Windows\MEMORY.DMP` (routinely 1-8 GB, sized to RAM), `C:\Windows\Minidump\*.dmp`, or `C:\Windows\LiveKernelReports`. On a machine that has blue-screened, MEMORY.DMP alone is usually the single largest reclaimable file on the drive. (2) **Thumbnail/icon cache** - `grep -rniE "thumbcache|iconcache_|FNTCACHE"` returns only `Services/AppIconService.cs:27` which is SysManager's OWN icon cache folder, not Windows'. `%LocalAppData%\Microsoft\Windows\Explorer\thumbcache_*.db` is commonly hundreds of MB and its corruption is the classic cause of "my folder thumbnails are wrong/blank" - a symptom the persona actually reports. (3) **WinSxS component store** - `grep -rn "StartComponentCleanup|AnalyzeComponentStore|ResetBase"` across .cs/.xaml/.md returns nothing; `winsxs` appears only in scanner EXCLUSION lists (`Services/DiskAnalyzerService.cs:23`, `Services/DuplicateFileService.cs:35`, `Services/LargeFileScanner.cs:23`). WinSxS routinely holds 5-10 GB of superseded components. The app already runs DISM for a different purpose (`ViewModels/CleanupViewModel.cs:468` runs `DISM.exe /Online /Cleanup-Image /RestoreHealth`), so the seam and the elevation/console plumbing exist.

## Proposed solution
[...]

PR #2246, opened 2026-09-11
feat: say when administrator rights are what Deep Cleanup is missing

laurentiu021:
Prerequisite for #1577. `feat:` -> v1.92.0.

### The gap

Deep Cleanup had **no elevation affordance at all**. Grepping the view and the view model for `admin|IsElevated|administrator` returned exactly one hit, and it was prose about the large-files finder being read-only.

Five of its nineteen buckets live under `%WinDir%`:

- Windows Update download cache
- Delivery Optimization cache
- Windows Installer patch cache
- `Windows\Temp`
- Prefetch

Unelevated, those files were scanned, sized, listed — and then reported through `CleanupCategory.SkippedCount` as `"N files · M skipped"`, with nothing anywhere saying that administrator rights were the reason. Those are among the **largest** items on the list, so the tab could look like it had done its job while leaving the biggest wins untouched. The one thing the user could act on was the one thing it did not say.

### The fix is the contract, not a new idea

The golden admin-control contract already specifies this: a privileged page explains WHY admin is needed and WHAT unlocks, at the top of the page. So: the shared `AdminBanner` in the documented position, plus the two members it binds ambiently (`IsElevated`, `RelaunchAsAdminCommand`).

Two details worth stating:

- `Margin="0,12,0,0"`, not a 28 gutter — Deep Cleanup is one of the three **scrolling pages**, whose `StackPanel` already carries the gutter. Using the per-section value would have double-indented it.
- `IsElevated` is read **synchronously in the constructor, before `InitializeAsync`**, so the banner does not flicker from "needs administrator" to "running as administrator" after the page has painted.

### No new guard, and I checked rather than assumed

`TheElevationBanner_LivesInOneControl_AndItsHostsExposeWhatItBinds` already covers exactly this failure — the banner binds ambiently, so a host missing either member renders a banner stuck in one state or a dead button, with no compiler error and nothing visible on a machine that never launches the app. Rather than trust that, I removed each half:

```
baseline: GREEN
mutation 1 IsElevated removed        -> the project no longer compiles (caught by the COMPILER)
mutation 2 RelaunchAsAdmin removed   -> RED  "DeepCleanupView — its view model exposes no
                                             RelaunchAsAdminCommand, so the button does nothing"
2 of 2 halves are caught
```

View model restored byte-for-byte. Recorded honestly that one half is a compile break rather than a guard hit — that is still caught, just by a different mechanism.

### Verification

- App and unit projects — **0 errors, 0 warnings**
- Unit suite **5480 passed**, 0 failed
- `dotnet format --verify-no-changes` clean on both projects
- Leak scan of the staged diff against all 43 patterns — 0 hits
- Docs in this PR: README Deep Cleanup section, CHANGELOG `[1.92.0]` with both an Added and a Fixed entry, SECURITY -> 1.92.x, csproj -> 1.92.0

### Why this is a prerequisite

#1577 wants two new buckets — blue-screen memory dumps and the Explorer thumbnail cache — that need admin to delete, and its risk note names this gap as what blocks them: without a banner they would "silently land in the existing SkippedCount path". Now they would not.

### Deferred to the secondary workstation

The banner is above the fold on a scrolling page, so it competes for space with the existing accent-coloured safety note directly beneath it. Two stacked notices at the top of one page is the kind of thing that reads fine in markup and heavy on screen — worth a look.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,24 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.92.0] - 2026-09-11
+
+**Deep Cleanup now tells you when administrator rights are the thing standing in the way.** Five of its
+categories live in the Windows folder — the Windows Update download cache, Delivery Optimization, the
+Installer patch cache, `Windows\Temp` and Prefetch — and without administrator they were counted, listed,
+and then quietly reported as "skipped" with no reason given. Those are among the largest items on the
+list, so the tab could look like it had done its job while leaving the biggest wins untouched. It now
+says so at the top of the page, in the same place every other tab does, with a button to restart elevated.
+
+### Added
+
+- **Elevation notice on Deep Cleanup**, naming the five categories that need administrator and what
+  becomes deletable once you restart elevated.
+
+### Fixed
+
+- Items skipped for lack of permission no longer look like items that were simply not there.
+
 ## [1.91.0] - 2026-09-11
 
 **Every tab whose output is a list can now save it to a file.** Five more got an **Export CSV** button —
--- a/README.md
+++ b/README.md
@@ -447,6 +447,11 @@ a confirmation before it runs:
 ### Deep Cleanup
 - **Scan-first**: every category is discovered with size + file count
   before a single byte is deleted. You pick what goes.
+- **Says when administrator rights are what's stopping it.** Five of the buckets live in
+  the Windows folder — the Windows Update download cache, Delivery Optimization, the
+  Installer patch cache, `Windows\Temp` and Prefetch — and without administrator they are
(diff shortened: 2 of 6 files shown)


laurentiu021:
Merging. Blocking job and UI suite both green (5480 and 136 passed).

The integration job reads `success` on the check row but its summary says `Failed! failed: 1` — the `continue-on-error` masking again. The failure is `PowerShellRunnerTests.RunAsync_SupportsCancellation` at 33.82s, which is #2206 and **not** this PR: nothing here touches a PowerShell path.

Worth stating that it made me reopen #2206 and withdraw my own conclusion. I had closed it as fixed by #2236 with the reopen criterion "this message on a run where the ping stress tests are behaving" — and on this run they were: 8.62s, 8.00s, 7.04s and 14.41s, whole suite 200s, against the 217s/283s that motivated the closure. So the ping drain was a real trigger but not the cause, and I closed on one demonstrated trigger. Detail on the issue.

Issue #2176, opened 2026-09-08
[Enhancement]: Deep Cleanup — give ScanAsync its roots, so its logic can be asserted at all


laurentiu021:
## Problem

`DeepCleanupService.ScanAsync` decides what to clean by asking the machine directly. It reads
`Environment.GetFolderPath` for five special folders, enumerates every drive root looking for `Steam`,
`SteamLibrary\steamapps\shadercache` and `Riot Games\...\Logs`, and walks each with
`Directory.EnumerateFiles`. There is no parameter, interface or constructor argument that redirects any
of it.

So its LOGIC cannot be tested. Everything the scan decides — which categories exist, how sizes
aggregate, that `windows.old` is never selected by default, that an empty category is not selected, that
a junction at a scan root is not followed — is only assertable against whatever happens to be on the
machine running the tests. Two consequences:

1. **The assertions are weak by necessity.** They can only check shapes that hold on any machine: the
   name is non-empty, the size is non-negative, the list contains something matching "Recycle". A test
   cannot say "a category over a folder holding 3 files of 10 bytes reports 3 files and 30 bytes",
   because it cannot create that folder anywhere the scan will look.
2. **CI asserts less than a developer's machine, and neither asserts much.** A fresh `windows-latest`
   runner has almost nothing in those caches, so most categories come back empty there; a real machine
   has content but unpredictable content. Whatever passes, passes for reasons no one chose.

#2167 fixed the COST of this — the class scanned the machine 27 times and took 23 minutes on a real
workstation; one shared scan brought it to 2.4. This is the other half: the scan is still
non-deterministic, and the shared fixture makes that more visible rather than less, since 27 tests now
depend on one scan of one machine.

## Proposed solution

Give the scan its roots instead of letting it find them.

```csharp
public DeepCleanupService(ICleanupRoots? roots = null)
```

where the default implementation is exactly today's behaviour — the five special folders plus the
drive-root probe — and a test can supply a temporary directory tree it built itself. That keeps
production wiring untouched (the parameterless constructor still works everywhere it is used) while
making the interesting assertions possible:

- a category over a known tree reports the exact file count and byte total
- `OlderThan` keeps a file one day younger than the cutoff and deletes one a day older
- a junction planted at a scan ROOT is not descended into, asserted on a tree the test owns rather than
  hoping the environment can create one (two current tests bail out with `return` when `mklink` fails,
  which is a silent skip)
[...]

PR #2248, opened 2026-09-11
feat: find the biggest file on a PC that has blue-screened

### LINKED ISSUE #1577: [Enhancement]: Cleanup / Deep Cleanup - Deep Cleanup misses the three biggest Windows space hogs: WinSxS component store, kernel memory dumps,...

laurentiu021:
Closes #1577 (partially — the WinSxS component store half is deliberately not here; see below).

## What this adds

Two buckets, and the one schema change both of them needed.

**Blue-screen memory dumps** — `MEMORY.DMP`, `Windows\Minidump`, `Windows\LiveKernelReports`. `MEMORY.DMP` is sized to installed RAM, so on a machine that has crashed it is routinely the single largest reclaimable file on the drive, and nothing in the app looked at it. It follows the `Windows.old` precedent: offered with its size and file count, `IsDestructiveHint`, never pre-selected. Those dumps are the only record of why the machine crashed, and the description says so.

**Explorer thumbnail & icon cache** — `thumbcache_*.db` and `iconcache_*.db`. Windows rebuilds them, and clearing them is the standard fix for thumbnails that come up blank or show the wrong picture.

## Why the schema had to change

Neither bucket fits "every file under these directories".

`MEMORY.DMP` is a **file**, sitting directly in `%WinDir%`. The scan only considered paths `Directory.Exists` agreed with, so naming it would have been silently dropped — and naming its parent instead would walk all of Windows. A `Def` path may now name a file.

The Explorer folder is **Explorer's own working directory**. Alongside the rebuildable caches it holds `AutomaticDestinations` — the jump lists that are the user's recent-files history. So a bucket may now restrict itself to files matching a wildcard, via `FileSystemName.MatchesSimpleExpression` so a pattern means here what it means to `Directory.EnumerateFiles`.

The restriction is carried on the `CleanupCategory`, mirroring `OlderThan`, **not** looked up from the definitions at clean time. `Clean` is handed categories by the caller and walks their `Paths`; a filter it could not see would produce the worst available outcome — an honest size on screen and a delete that took the whole folder. Both sides now go through one helper, `EnumerateTargets`, so the set a bucket reports is by construction the set it deletes. A filtered bucket also leaves its emptied folders in place: it owns files, not the folder.

## The infinite loop this feature made reachable

The proof mutation that replaced the filtered call with the raw walk did not fail — it ran for **900 seconds** on a class that takes 0.24. That is a real defect in the walker, not a test artifact.

`EnumerateFiles` caught a throwing `MoveNext` and `continue`d, a shape written for "an entry becomes unreadable mid-enumeration". Measured with a standalone probe rather than reasoned about:

| path | `EnumerateFiles(...)` | `GetEnumerator()` | `MoveNext()` |
|---|---|---|---|
| missing directory | throws `DirectoryNotFoundException` | — | — |
| directory the user may not list | throws `UnauthorizedAccessException` | — | — |
| **a file** | returns | returns | throws `IOException` **10 times out of 10, never advancing** |

The first two throw from the call, where the existing `catch { continue; }` sits on the outer loop and continuing means "take the next directory off the stack" — real progress, and why an unelevated scan of Prefetch reports "skipped" rather than hanging. The third reaches the inner `while (true)`, where `continue` re-threw forever and pinned a core in a loop the cancellation token is never checked inside.

It stayed unreachable for exactly as long as both callers filtered their paths through `Directory.Exists`. This PR is what makes a file path legitimate, so it now abandons the directory instead. Minimal diff: two `continue`s became `break`s, and the comment says what was measured.

## Tests

11 behaviour tests over a temp tree the test owns (`TempCleanupRoots`, the `ICleanupRoots` seam from #2176), asserting Scan and Clean filter **independently** — they are two code paths and a filter in one of them is a bucket that lies. Plus two guards.

Red-before, green-after: 10 of the 11 failed with "no category named …" against the unfixed code; the 11th passed and is there to pin the behaviour every *unfiltered* bucket relies on (emptied folders still get removed).

Every guard assertion was proven red by mutation, each reverted byte-for-byte:

| mutation | what went red |
|---|---|
| `Matches` returns true for everything | 2 scan tests + the Clean jump-list test |
| Scan stops copying the filter onto the category | guard + 2 bucket tests |
| Clean stops reading the filter off the category | guard + 2 bucket tests |
| Scan walks files directly, past the filter | guard (`raw EnumerateFiles walk 3 times`) |
| the dir-delete gate inverted | both the filtered **and** the unfiltered folder test |
| the retry restored (`break` → `continue`) | `TheCleanupWalk_AbandonsADirectoryWhoseEnumeratorThrows` |

The walker guard is a source guard on purpose: a spin fails no assertion, it just never returns, so nothing else would catch its return.

`EveryTempTreeWalkerCall_PassesBothExtractionExclusions`' vacuity floor moves 7 → 6. Two direct walks became one; the population shrank, the pattern did not stop matching, and the comment records which.

## Verification

- Unit suite: **5493 passed, 0 failed** (5480 → 5493: 11 behaviour tests + 2 guards).
- `dotnet format --verify-no-changes`: clean on both projects (exit 0, captured without a pipe masking it).
- Full local pre-commit scan against all 43 current patterns: 3 pre-existing matches, every one of them a legitimate product string that predates this branch. Nothing from this diff.
- Docs in this same PR: README (the six `%WinDir%` buckets, and both new buckets with what is *not* deleted), ARCHITECTURE (the `FilePatterns` contract under the Deep Cleanup guardrails), CHANGELOG, and the elevation banner copy — five → six, the dumps being the sixth.

## Deliberately not here

#1577 also asks for the **WinSxS component store**, and its own text is right that it must not be a file bucket — deleting WinSxS files directly corrupts servicing. That half is a `DISM /Online /Cleanup-Image /AnalyzeComponentStore` report followed by an opt-in `/StartComponentCleanup`, which belongs next to the existing SFC/DISM controls in Quick Cleanup rather than in this bucket list. Separate change, separate risk profile: a 10–30 minute operation that cannot be cancelled cleanly mid-servicing.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -859,7 +859,15 @@ tab cannot ship a refresh or cancel button the keyboard cannot reach.
 - Scan first, clean second. Every category is opt-in and shows its size.
 - Never touches browsers, passwords, registry, active drivers, Program
   Files, or actual game files in `steamapps\common`.
-- Windows.old is tagged **Irreversible** and never selected by default.
+- Windows.old is tagged **Irreversible** and never selected by default, as are the
+  blue-screen memory dumps: they are the only record of why a machine crashed.
+- A bucket may restrict itself to files matching a wildcard (`FilePatterns`), and the
+  restriction is carried on the `CleanupCategory` rather than looked up at clean time —
+  the cleaner is handed categories and walks their `Paths`, so a filter it could not see
+  would show an honest size and delete the whole folder. Two buckets need it: the dumps
+  (`*.dmp`, among `.etl` traces) and the Explorer thumbnail cache, whose folder also holds
+  the jump lists that are the user's recent-files history. A filtered bucket also leaves
+  its emptied folders in place, because it owns files and not the folder.
 - Large files finder has no delete action, even with admin rights.
 
 ## Logging
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,33 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.93.0] - 2026-09-11
+
+**Deep Cleanup can now find the biggest file on a PC that has blue-screened.** When Windows crashes it
+writes out `MEMORY.DMP`, sized to how much RAM the machine has — routinely several gigabytes, and often
+the single largest reclaimable file on the drive. Deep Cleanup had no bucket for it, nor for the
+`Minidump` and `LiveKernelReports` folders beside it. It now does, and it is deliberately never ticked
+for you: those dumps are the only record of why the machine crashed, so removing them ends any
+investigation into it. The second addition is the thumbnail cache — the previews Windows keeps so folders
+open quickly, whose clearing is the standard fix for thumbnails that come up blank or show the wrong
+picture.
+
+### Added
+
+- **Blue-screen memory dumps** as a bucket of its own: `MEMORY.DMP`, `Windows\Minidump` and
+  `Windows\LiveKernelReports`. Offered with its size and file count like every other bucket, never
+  pre-selected, and it needs administrator to delete — which the tab now says.
+- **Explorer thumbnail & icon cache**, scoped to exactly `thumbcache_*.db` and `iconcache_*.db`. That
+  folder is Explorer's own working directory and also holds the jump lists that are your recent-files
+  history, so nothing else in it is counted, and nothing else in it is deleted.
+
+### Changed
+
+- A cleanup bucket can now name a single file rather than a folder, and can restrict itself to files
+  matching a pattern. Both the scan and the delete read the same restriction, so the size shown is by
+  construction the size removed.
+- The Deep Cleanup elevation notice names six categories instead of five, the dumps being the sixth.
+
 ## [1.92.0] - 2026-09-11
 
 **Deep Cleanup now tells you when administrator rights are the thing standing in the way.** Five of its
--- a/README.md
+++ b/README.md
@@ -447,15 +447,23 @@ a confirmation before it runs:
 ### Deep Cleanup
 - **Scan-first**: every category is discovered with size + file count
   before a single byte is deleted. You pick what goes.
-- **Says when administrator rights are what's stopping it.** Five of the buckets live in
+- **Says when administrator rights are what's stopping it.** Six of the buckets live in
   the Windows folder — the Windows Update download cache, Delivery Optimization, the
-  Installer patch cache, `Windows\Temp` and Prefetch — and without administrator they are
-  still scanned and counted but cannot be deleted. They used to show as "skipped" with no
-  reason given; the tab now says so at the top, and offers to restart elevated
+  Installer patch cache, `Windows\Temp`, Prefetch and the blue-screen memory dumps — and
+  without administrator they are still scanned and counted but cannot be deleted. They used
+  to show as "skipped" with no reason given; the tab now says so at the top, and offers to
+  restart elevated
 - **System buckets**: NVIDIA / AMD / Intel installer leftovers, Windows
   Update cache, Delivery Optimization cache, Windows Installer patch
   cache, TEMP, Prefetch, crash dumps, old CBS logs, DirectX shader cache,
   Recycle Bin on every drive.
+- **The two biggest wins on a machine that has crashed**, both new and both scoped to
+  exactly the files they name: the **blue-screen memory dumps** (`MEMORY.DMP` is sized to
+  your RAM, so it is routinely gigabytes — never ticked for you, because deleting it ends
+  any investigation into the crash) and the **Explorer thumbnail & icon cache**, which
+  Windows rebuilds and whose clearing is the standard fix for blank or wrong thumbnails.
+  Its folder also holds your recent-files jump lists, so only `thumbcache_*.db` and
+  `iconcache_*.db` are counted, and only those are deleted.
 - **Gamer buckets** — launcher *caches only*, never game files or logins:
   Steam (appcache, htmlcache, depotcache, shader cache), Epic Games
   Launcher, Battle.net, Riot / League of Legends, GOG Galaxy, EA Desktop.
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -12,8 +12,8 @@ older build, the first step is usually to update.
 
 | Version  | Supported          |
 | -------- | ------------------ |
-| 1.92.x   | :white_check_mark: |
-| < 1.92   | :x:                |
+| 1.93.x   | :white_check_mark: |
+| < 1.93   | :x:                |
 
 The supported line is always the newest minor on the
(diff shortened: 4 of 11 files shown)


laurentiu021:
Blocking job green (5493 unit tests passed), CodeQL green, UI suite green (136 passed).

The integration job's check row reads `pass`, but its own summary says `failed: 1` of 631 — the job is `continue-on-error`, which masks a failure at the row level. The failure is `PowerShellRunnerTests.RunAsync_SupportsCancellation` at 32.03 s, which is #2206: the fifth reading, all five clustering on the 30-second `Start-Sleep` with nothing in between. This branch touches Deep Cleanup's file walk and nothing on any PowerShell path, so it is not a contributor. Recorded on #2206 rather than here.

Merging.

Issue #2201, opened 2026-09-09
[Bug]: Cross-app — 66 tests assert on a StatusMessage that a fire-and-forget constructor init can overwrite mid-test


laurentiu021:
## Problem

Most view-models call `InitializeAsync(InitAsync)` from their constructor and forget the task. When that
init reaches an `await` before it writes `StatusMessage`, its continuation is still pending while a test
runs — and the test's own `await SomeCommand.ExecuteAsync(...)` is a yield point that lets the
continuation land. The init's message then overwrites whatever the command reported, and the assertion
fails on a string neither the test nor the command produced.

This is not hypothetical. It cost a red CI run on #2199, a pull request that touched only the theme stack:

```
PerformanceViewModelTests.ApplyPowerPlan_WhenSystemModificationLocked_BailsAtGuard
  Assert.Contains() Failure: Sub-string not found
  String:    "Reading performance settings…"     <- PerformanceViewModel.RefreshAsync, from InitAsync
  Not found: "already running"                   <- the lock guard the test is asserting
```

Two tests were fixed in #2200. This issue is the exposure that remains.

## Measured, on current source

| | |
|---|---|
| view-models calling `InitializeAsync` from their constructor | **40** |
| test methods asserting on a `StatusMessage` | **117** |
| of those, never awaiting `InitializationComplete` | **66** |
| async methods writing `StatusMessage` inside those view-models | **91**, across **34** view-models |

The 66 span `AudioMixerViewModelTests`, `BandwidthMonitorViewModelTests`, `CleanupViewModelTests`,
`CpuAffinityViewModelTests`, `DnsHostsViewModelTests` (12 methods), `PrivacyViewModelTests`,
`ServicesViewModelTests`, `SystemHealthViewModelTests` and others.

## Why there is no guard yet

A rule of the shape "a test asserting on `StatusMessage` must await `InitializationComplete`" would land
with dozens of findings that are **correct as written**:

- `CleanupViewModelTests.Constructor_InitialStatusMessageIsEmpty` and
  `DashboardViewModelTests.Constructor_StatusMessageEmpty` assert the pre-initialisation state deliberately —
  awaiting init is exactly what they must not do.
- Several view-models' init paths never reach a `StatusMessage` write at all, so their tests cannot race.
- `PerformanceViewModelTests.RestoreAll_WhenSystemModificationLocked_BailsAtGuard` awaits through a
  helper, so a body-scoped scan reports it as a violation when it is safe.

[...]

PR #2257, opened 2026-09-11
fix: a cancelled PowerShell query was reported as a successful empty result

### LINKED ISSUE #2206: [Bug]: CI — the integration cancellation test ran a 30s script to completion once, and the obvious cause is refuted

laurentiu021:
Closes #2206.

Found by doing what that issue's own plan asked for — **step 2, make the test say more** — rather than by patching the runner. The instrumentation named the root cause on the first local run.

## Root cause

`RunOnLeasedRunspaceAsync` reported cancellation only when `EndInvoke` threw `PipelineStoppedException`. **A stopped pipeline does not always throw.**

Opening the runspace takes a few hundred milliseconds. A cancel landing in that window means the token is already cancelled when `cancellationToken.Register(...)` runs — and `Register` invokes the callback **synchronously** in that case, so `ps.Stop()` lands on a `NotStarted` instance. `BeginInvoke`/`EndInvoke` then complete without throwing *and without running the script*, and the method returns an empty collection.

Reproduced locally on the first attempt, with the new instrumentation:

```
elapsed 00:00:00.3337465; token fired at 318 ms; exception NONE; 0 line(s) captured
```

The caller was handed an empty result set and no indication that anything had been cancelled. For a query like "which DNS servers am I using?" or "what did Defender find?", an empty collection is **indistinguishable from a real answer of "none"** — so the tab renders an empty state instead of saying it was cancelled. 31 view-models have an `OperationCanceledException` branch and the services deliberately let it propagate; this was the one path that never raised it.

The same hole explains the CI shape from the other end: a `Stop()` that fails to interrupt a *running* pipeline, which then finishes normally 30 seconds later — again returning without an exception. That is why all five CI failures sat on the script's own duration (31.12, 30.565, 30.664, 33.82, 32.03 s) with nothing in between.

## The fix, and why it differs from its sibling

One post-await check. Notably it is the **opposite** choice from `RunProcessAsync`, which on the same race deliberately lets completion win — *"callers receive the real exit code instead of a false cancellation"*.

That is right there and wrong here, and the difference is what the method returns: an exit code from a process that finished is a true and complete answer worth preserving, whereas an empty collection from a script that never ran is not an answer at all. There is nothing to preserve by staying silent. The code says so, so the next reader does not have to reconcile the two by guessing.

## The two paths now say which one happened

Both raise `OperationCanceledException`, so before this a slow cancellation could not be told apart from a stop that never interrupted anything — which is precisely why the bug needed five CI failures to pin down. Two distinct messages now:

| message | means |
|---|---|
| "was stopped by cancellation" | `ps.Stop()` interrupted the pipeline. If elapsed is also long, only the bound failed. |
| "was not interrupted by cancellation; it completed on its own" | the script never started, or ran to its natural end. The CI failure shape. |

I found and fixed a flaw in my own diagnostic here: with the post-await throw in place, "OCE + slow elapsed" no longer distinguishes the two, so the message is doing the work the exception type used to.

## Three tests, and why only one is the regression test

**`CancelledWhileTheRunspaceIsOpening`** — deterministic. The `openRunspace` seam blocks until the test has cancelled, so the token is *guaranteed* already cancelled when `Register` runs. No timing, no sleeping, same sequence on every machine. Red before the fix with *"reported NO EXCEPTION"*, green after, **in a full-class run**. It asserts the exception **message**, so it cannot pass on the path that already worked.

**`CancellingARunningPipeline`** — the other path, cancelling a pipeline that is genuinely executing. The script is `while ($true) { [System.Threading.Thread]::Sleep(50) }` and not `Start-Sleep`, because of trap #1 from the issue: `Start-Sleep` lives in `Microsoft.PowerShell.Utility`, which `CreateDefault2()` does not load, so on a dev box it fails instantly and a cancellation that never happened reads as a fast pass. That trap cost a whole round of investigation, so the test asserts the probe **actually blocked** before asserting anything else.

**`SupportsCancellation`** — kept as the CI canary, now reporting elapsed, when the token fired, the exception type *and* its message.

### Why the canary is not the regression test

Its coverage depends on execution order, which I only discovered by mutating:

| run | runspace open | path taken |
|---|---|---|
| alone, cold process | ~300 ms (assembly loading dominates) | cancel lands during the lease → the swallowed path |
| after any other test in the class | ~20 ms | pipeline already running → the throwing path |

**Measured: removing the fix leaves the whole class GREEN, while running that test alone against the same code fails with "exception NONE".** A test whose coverage depends on execution order is not a regression test, so the deterministic one exists.

## Mutations

| mutation | result |
|---|---|
| the post-await throw removed | `CancelledWhileTheRunspaceIsOpening` red — *"reported NO EXCEPTION"* |
| the translation arm downgraded to a bare `rethrow` | both cancellation tests red with `PipelineStoppedException` — proving the two arms cover different paths |
| the post-await condition inverted | **13 red**, proving the condition is what scopes the throw to cancelled runs |

Two earlier mutation attempts did not compile — `when (false)` is rejected as a constant filter (CS8360) and `if (true)` makes the `return` unreachable (CS0162) — so each was rewritten into a shape the compiler accepts. Worth recording, since the obvious mutation is not always available.

## Verification

- Unit suite: **5551 passed, 0 failed**.
- `PowerShellRunnerTests` locally: **27 passed** (25 → 27). Its runspaces are in-process and its one process-spawning test pings loopback by design, so nothing here needs elevation or the network.
- `dotnet format --verify-no-changes`: clean on all four projects.
- Full local pre-commit scan against all 43 current patterns: **0 matches in the added lines**, and no machine name in the diff.
- CHANGELOG in plain language — the user-visible symptom is "Cancel left the tab looking like it had finished with nothing to show". `SECURITY.md` needs no edit: this is a patch, and only the minor line is tracked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,12 @@ whichever moment it lands in.
 - A cancelled PowerShell query is no longer reported to the app as a successful run that returned nothing.
   It only reported cancellation when the engine was interrupted mid-script; a cancel that arrived before the
   script started, or one that failed to interrupt a script already running, was silently discarded.
-- The two cancellation paths now say which one happened, so a failure of this kind can be diagnosed from its
+- **Cancelling now works the same way with administrator rights as without.** The two are not the same
+  underneath — running elevated hands the work to a separate Windows PowerShell process — and the cancelled
+  result came back in a different form from there that the app did not recognise. So on an elevated session
+  a cancelled query surfaced as a raw engine error rather than as a cancellation. This is the half that
+  could only ever be seen on a machine running as administrator.
+- The cancellation paths now say which one happened, so a failure of this kind can be diagnosed from its
   message instead of from a stopwatch. That is what this bug needed five CI failures to work out.
 
 ## [1.96.0] - 2026-09-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,23 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.96.1] - 2026-09-11
+
+**Pressing Cancel could leave a tab looking like it had finished successfully with nothing to show.** When a
+PowerShell query was cancelled in a narrow window — while its engine was still starting up — the query never
+ran, but nothing told the tab that. It received an empty list of results, which for a question like "which
+DNS servers am I using?" or "what did Defender find?" looks exactly like a real answer of "none". The tab
+would then show an empty state instead of saying it had been cancelled. Cancelling now always reports itself,
+whichever moment it lands in.
+
+### Fixed
+
+- A cancelled PowerShell query is no longer reported to the app as a successful run that returned nothing.
+  It only reported cancellation when the engine was interrupted mid-script; a cancel that arrived before the
+  script started, or one that failed to interrupt a script already running, was silently discarded.
+- The two cancellation paths now say which one happened, so a failure of this kind can be diagnosed from its
+  message instead of from a stopwatch. That is what this bug needed five CI failures to work out.
+
 ## [1.96.0] - 2026-09-11
 
 **The app now opens showing something you can actually use.** Every sidebar group used to start collapsed,
--- a/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
+++ b/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
@@ -468,17 +468,194 @@ public class PowerShellRunnerTests
         Assert.True(gotError);
     }
 
+    /// <summary>
+    /// Cancelling a running script must stop it, and must surface that as
+    /// <see cref="OperationCanceledException"/> so a caller's cancel branch fires.
+    /// </summary>
+    /// <remarks>
+    /// This test has failed on CI five times with the elapsed time sitting on the script's own duration —
+    /// 31.12, 30.565, 30.664, 33.82, 32.03 seconds against a 30-second sleep, and nothing in between. So the
+    /// failure is all-or-nothing: either <c>ps.Stop()</c> takes effect within a few hundred milliseconds or
+    /// the pipeline runs to its natural end (#2206). The obvious cause — the cancellation callback landing on
+    /// a not-yet-invoked <c>PowerShell</c> instance — was refuted with a harness that mutated the runner three
+    /// ways and cancelled in ~0.35 s every time, so no speculative fix was shipped.
+    ///
+    /// <para><b>What this instrumentation is for.</b> The failure message used to be one number, which cannot
+    /// tell the two live hypotheses apart. The runner's contract makes them distinguishable:
+    /// <c>ps.Stop()</c> makes <c>EndInvoke</c> throw <c>PipelineStoppedException</c>, which
+    /// <c>RunOnLeasedRunspaceAsync</c> translates to <c>OperationCanceledException</c>. So if Stop never
+    /// takes effect the script completes, <c>EndInvoke</c> returns normally and <b>no exception is thrown at
+    /// all</b>.</para>
+    ///
+    /// <list type="bullet">
+    /// <item>slow + <c>OperationCanceledException</c> → Stop worked and the 5-second bound is simply tight.</item>
+    /// <item>slow + <b>no exception</b> → Stop had no effect; the pipeline ran to completion. A user pressing
+    /// Cancel would wait out the whole operation.</item>
+    /// <item>token never fired → the <c>CancellationTokenSource</c> itself was starved, which on a loaded
+    /// runner is a test-environment fact rather than a product defect.</item>
+    /// </list>
+    ///
+    /// <para>The token's own firing time is captured through a SECOND registration on the same token rather
+    /// than by instrumenting the runner: it needs no production change, and it separates "the token fired
+    /// late" from "the token fired on time and was ignored", which is the whole question.</para>
+    ///
+    /// <para>Error lines are captured because of a trap this bug already produced twice: on a dev box
+    /// <c>Start-Sleep</c> does not exist in this runner's runspace — <c>CreateDefault2()</c> loads
+    /// <c>Microsoft.PowerShell.Core</c> only — so the script errors instantly, returns normally, and a
+    /// cancellation that never happened reads as a fast success. If that is what is happening, the captured
+    /// error says so instead of leaving the next person to rediscover it.</para>
+    /// </remarks>
     [Fact]
     public async Task RunAsync_SupportsCancellation()
     {
         var runner = new PowerShellRunner();
-        using var cts = new CancellationTokenSource(TimeSpan.FromMilliseconds(300));
+        var lines = new System.Collections.Concurrent.ConcurrentQueue<Models.PowerShellLine>();
+        runner.LineReceived += lines.Enqueue;
+
+        using var cts = new CancellationTokenSource();
         var sw = System.Diagnostics.Stopwatch.StartNew();
-        await Record.ExceptionAsync(async () =>
+
+        // Elapsed at the moment the token fired, from a registration of our own. -1 means it never fired.
+        var firedAtMs = -1.0;
+        using var observer = cts.Token.Register(() => firedAtMs = sw.Elapsed.TotalMilliseconds);
+        cts.CancelAfter(TimeSpan.FromMilliseconds(300));
+
+        var ex = await Record.ExceptionAsync(async () =>
             await runner.RunAsync("Start-Sleep -Seconds 30", cancellationToken: cts.Token));
         sw.Stop();
+
+        var errors = lines.Where(l => l.Kind == Models.OutputKind.Error).Select(l => l.Text).ToList();
+        var diagnosis =
+            $"elapsed {sw.Elapsed}; token fired at {(firedAtMs < 0 ? "NEVER" : $"{firedAtMs:F0} ms")}; "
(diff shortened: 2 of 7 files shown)


laurentiu021:
## CI found the other half, and it is the actual reason this bug survived

The instrumentation added in the first commit did its job on the first CI run:

```
a cancelled running pipeline did not report cancellation.
  elapsed 00:00:02.2249351; token fired at 2224 ms;
  exception RemoteException (The pipeline has been stopped.)
```

**`RemoteException`, not `PipelineStoppedException`.** The translation arm named the in-process type only, so it never matched on CI.

Why: elevated, `LeaseRunspaceAsync` builds a `PowerShellProcessInstance(new Version(5, 1), …)` — an out-of-process Windows PowerShell 5.1 child reached over remoting. A stopped pipeline there comes back as a `RemoteException` carrying a *serialized* copy of the failure. GitHub's Windows runners run as administrator, so CI always took that branch; this machine runs unelevated and always took the in-process one.

That also settles the question the issue flagged as unexplained — *"CI clearly can run `Start-Sleep`, so the environments differ in a way that matters"*. It differs because elevated means out-of-process 5.1, which has the full module set, while the in-process `CreateDefault2()` runspace does not.

So the two halves of #2206 are:

| | unelevated (this machine) | elevated (CI) |
|---|---|---|
| cancel during runspace open | returns silently, **no exception** | same |
| cancel mid-pipeline | `PipelineStoppedException` → translated | **`RemoteException` → not matched, escapes raw** |

The first commit fixed the top row. This one fixes the bottom.

## The remote arm matches the type NAME, not the type

This is the part where the obvious implementation would have compiled, passed locally, and changed nothing on CI:

```csharp
(ex as RemoteException)?.SerializedRemoteException?.BaseObject is PipelineStoppedException   // always false
```

Remoting does not hand back the original exception object. It rehydrates a `PSObject` whose `TypeNames` read `Deserialized.System.Management.Automation.PipelineStoppedException`, so an `is` test against the real type never matches. Suffix-matching the type name is what actually holds — and it stays locale-independent, unlike matching `"The pipeline has been stopped."`, which a non-English Windows would fail silently.

## Tests: unit, not integration, deliberately

Producing a real remote failure needs an elevated host, so `PipelineStoppedDetectionTests` asserts the shapes directly — including building the `RemoteException` through its non-public constructor, and failing loudly if that shape ever changes rather than quietly asserting over a plain exception.

The point is that **the shape CI produced is now checked on any machine, at no cost.**

| mutation | what went red |
|---|---|
| the remote arm removed (the shipped defect) | `TheRemotedShape_IsRecognisedByItsDeserializedTypeName` |
| the inner-exception arm removed | `AWrappedOne_IsRecognised` |
| the predicate widened to accept everything | **two negative tests** — an unrelated remote failure must not be relabelled as cancellation, or a real script error during a cancelled run disappears |

## Also in this branch: #2201's step 1, which was "count before acting"

That issue records **66** tests racing a fire-and-forget init and explicitly says to count first rather than rush a fix. Doing the counting changed the answer three times:

| refinement | racy |
|---|---|
| the issue's body-scoped count | 66 |
| tighten bucket (a) to "asserts the message is still EMPTY" (four tests assert the opposite, which is the same race inverted) | 29 |
| trace the init ENTRY through its calls: 35 view-models → 22 that can write `StatusMessage` during init | 27 |
| resolve factory-level awaits — 7 of the 8 flagged files settle in their own `NewVm()` | **3** |

All three were in `DebloaterViewModelTests`, and they are fixed at the source with a settling `NewVm` overload that **all six** construction sites use. Six rather than three because the init also calls `Apps.ReplaceWith`, and every site adds its fixture immediately after constructing — a late init discards it, so a test asserting "nothing was removed" would pass for entirely the wrong reason.

`CleanupViewModel` is why the loose rule was not good enough: its init writes `TempSizeLabel` and `RecycleBinLabel` and never touches `StatusMessage`, so its eight tests cannot race. A guard shipping with those eight findings gets suppressed rather than fixed — which is the outcome #2201 warned about by name.

`EveryTestAssertingAStatusMessage_SettlesTheConstructorInitFirst` now holds the line at zero, with a three-name allowlist for the tests that assert the pre-init state deliberately. It needed one fix of its own, found by mutation: the name→body map kept only the first declaration, so a settle in **one** `NewVm` overload vouched for the other, and removing it from the one the racy tests call left the guard green. A helper counts as settling only when every overload does.

## Two product defects surfaced and filed, not fixed here

Settling the existing parameterless `NewVm` made two tests that had always passed start failing:

```
System.NullReferenceException
   at DebloaterService.ParsePackages(IEnumerable`1 objects)
   at DebloaterViewModel.RefreshAsync()
   at ViewModelBase.RunInitAsync(...)
```

- **#2258** — `RunInitAsync` catches six exception types and loses everything else. Nothing in production awaits `InitializationComplete`, so the exception becomes an unobserved task exception, which .NET Core ignores: no crash, no log, a half-initialised tab. 40 view-models fire an init this way. The fix is a broad catch at a fire-and-forget boundary, which this repo's rules forbid on sight — so it is your call, with three options costed on the issue rather than picked by me.
- **#2259** — `ParsePackages` is `public static` and documented for direct use, and dereferences its argument with no null guard. Not reachable from the real runner, which never returns null; reachable from anything else that calls it.

## Verification

- Unit suite: **5557 passed, 0 failed** (5551 → 5557: 5 predicate tests + 1 guard).
- `PowerShellRunnerTests` locally: **27 passed**.
- `dotnet format --verify-no-changes`: clean on all projects.
- Leak scan against all 43 patterns: **0 matches** in the added lines.

PR #2288, opened 2026-09-12
fix: Cancel was accepted, reported, and ignored if it arrived while the runspace was opening

### LINKED ISSUE #2286: [Bug]: the PowerShell cancellation test asserts a wall-clock bound on a race it documents as sometimes lost
## Problem

`PowerShellRunnerTests.RunAsync_SupportsCancellation` failed once on CI with a 22.5-second elapsed time against a 5-second bound:

```
failed SysManager.IntegrationTests.PowerShellRunnerTests.RunAsync_SupportsCancellation (22s 529ms)
  Cancellation took too long. elapsed 00:00:22.5266139; token fired at 390 ms;
  exception OperationCanceledException (The PowerShell pipeline was not interrupted by cancellation;
  it completed on its own.); 0 line(s) captured
```

The test's own diagnosis message is what makes this readable, and it says the right thing: the token fired on time (390 ms) and `ps.Stop()` **did not interrupt the pipeline**, so the script ran to its natural end — `BlockingScript` is a bounded 20-second loop, which matches the 22.5 s.

## The product is not wrong here

This is a known and documented hole, and the runner handles it. `PowerShellRunner.RunAsync`'s post-await check exists precisely for it, and its comment already describes this direction:

> Two ways to get here with the token cancelled, both measured rather than reasoned about: […] and on CI the opposite end of the same hole — **Stop failing to interrupt a running pipeline, which then finishes normally 30 seconds later.**

So the contract held: a cancelled run raised `OperationCanceledException`, with the message that says which of the two paths happened. A caller's cancel branch runs correctly. Nothing was reported as a success that wasn't one — which is the defect #2206 was about.

There is a real user-facing cost — pressing Cancel can mean waiting out the operation — but `ps.Stop()` is the only lever the runner has, and correctness is preserved either way.

## What is actually broken is the assertion

```csharp
Assert.True(sw.Elapsed < TimeSpan.FromSeconds(5), "Cancellation took too long. " + diagnosis + …);
```

That is a wall-clock bound on a race the same file documents as sometimes lost. When `Stop()` interrupts, the elapsed time is the runner's; when it does not, the elapsed time is **the script's**, and no runner change can bring it under five seconds. So the assertion is flaky by construction: it fails on the branch it knows about, and it fails with a message explaining why it was allowed to.

That makes it a broken test rather than a flake to re-run: it will keep firing at whatever rate the CI runner loses that race, and each time it will look like a regression in cancellation.

## Proposed solution

Split the contract from the timing, keeping both:

- **Contract, unconditional:** a cancelled run raises `OperationCanceledException`, and its message is one of the two known discriminators. This is deterministic and is the part every caller depends on.
- **Timing, conditional on the branch:** assert the 5-second bound **only when the message says the pipeline was stopped by cancellation** — that is, only when `Stop()` actually interrupted, which is the only case where elapsed time measures the runner rather than the script.

When Stop does not interrupt, the useful assertion is the one already below it: that the run still reports cancellation rather than an empty success. Bound the whole thing by the script's own length plus a margin, so a genuine hang still fails rather than passing as "Stop didn't interrupt".

Both branches keep the existing diagnosis string, which is the reason this failure was legible at all.

## Not to be done

Do not raise the bound to 25 seconds. That would make the assertion pass by measuring nothing — 22.5 s is the script's length, so any bound above it accepts a cancellation that did nothing, which is the case the test was written to detect.

## Affected area

`SysManager.IntegrationTests/PowerShellRunnerTests.cs` · non-blocking integration job

## Notes

Surfaced by reading the job log rather than the check: the integration job is `continue-on-error`, so `gh pr checks` printed **pass** while the log said `Failed! failed: 1` of 641. First occurrence in the last four runs — it passed on the three before it, on unrelated branches, which is consistent with a load-dependent race rather than a code change. #2283's step now names this shape on the unit job; the integration job already had it.

laurentiu021:
Closes #2286. **I filed that issue as a flaky test and it was wrong** — the test was right and its own diagnosis message was the answer. Correcting that is most of what this change is.

## The defect

`RunAsync` registers `ps.Stop()` on the token *before* invoking:

```csharp
using var reg = cancellationToken.Register(() => { try { ps.Stop(); } catch (InvalidOperationException) { } });

var task = Task.Factory.FromAsync(ps.BeginInvoke<PSObject, PSObject>(null, output), ar => ps.EndInvoke(ar));
```

If the token fires while the instance is still `NotStarted` — which it is for the few hundred milliseconds a runspace lease and open take — `Stop()` throws `InvalidOperationException`, the catch swallows it, and `BeginInvoke` then starts a pipeline **nothing has asked to stop**.

The post-await check still reports `OperationCanceledException`, so the caller's cancel branch runs and the tab says it stopped. The script runs to the end anyway, in the background.

## Why I first read this as a test problem, and what changed my mind

CI showed a cancel at 390 ms against a script that loops on 50 ms sleeps, and the run finishing its whole 20 seconds reporting *"not interrupted … completed on its own"*. My first reading: the 5-second assertion was a wall-clock bound on a race PowerShell does not promise to win, so the test was at fault.

It is not. The blocking-call test in the same file measures **that same loop shape cancelling in 2.2 s** — `ps.Stop()` interrupts a pipeline between statements, and a loop of 50 ms sleeps has a statement boundary every 50 ms. So the stop was not *refused*, it was **lost**. The assertion was correct and the message it printed named the cause; I had read "took too long" as a symptom of a loaded runner.

That matters beyond this fix: had I shipped the issue as filed, the timing assertion would have been relaxed and the defect would have gone quiet.

## The fix

Re-assert `Stop()` after `BeginInvoke`, where the instance will accept it. Idempotent — on an already-stopping pipeline it raises the same `InvalidOperationException` this swallows, and on a running one it does what the registration meant to do the first time.

## A deterministic reproduction already existed

I built a simulation before noticing that. The `openRunspace` seam holds the open until the test has cancelled, so the token is *guaranteed* cancelled before `BeginInvoke` — precisely the window. The existing test using it runs `1 + 1`, where "the script ran anyway" is invisible.

The new test runs the same seam with `BlockingScript`, which separates them outright:

| state | result |
|---|---|
| re-assert **removed** | **RED, 20.5 s** — the script ran in full: the CI symptom exactly |
| re-assert **present** | **GREEN, 0.46 s** |
| whole class, either way | 29 tests, and only that one differs |

No sleeping to arrange the race, and no dependence on execution order — which is what the two older cancellation tests suffer from, as their own docs say.

## Two existing tests corrected, both because they had encoded the hole

- `RunAsync_CancelledWhileTheRunspaceIsOpening_ReportsCancellation` asserted the message `"completed on its own"`. That was the honest description of the defect. With the window closed, which message comes back is **genuinely racy for a trivial script** — the stop and `1 + 1` finishing are in a real race, and cancellation is correctly reported either way — so it now accepts either discriminator. Pinning the winner of a race is exactly what made the other test look flaky; the deterministic pin lives in the new test, where the script blocks.
- That test's doc claimed `BeginInvoke`/`EndInvoke` "complete without throwing **and without running the script**". They do not — the script runs. The claim was untestable against `1 + 1` and is measurably wrong against a blocking one. Corrected in place rather than deleted, since the reasoning around it is still right.

`RunAsync_SupportsCancellation` now asserts the **discriminator** rather than only the clock. For a looping script, "completed on its own" means the stop was lost — a defect, not a slow runner. The 5-second bound is **kept** as a backstop, below the script's own 20 seconds so a lost stop cannot pass it, with a message that says which of the two findings it is.

## Verification

**5576** unit tests pass, 0 failed. **29/29** `PowerShellRunnerTests`. Four projects build with **0 errors, 0 warnings**. `dotnet format --verify-no-changes` clean on all four. Leak scan: 43 terms over 370 diff lines, 0 hits. Ran after merging main in.

The integration suite as a whole is left to CI — parts of it touch the network, which is not run on this workstation.

## Scope note

The user-visible effect is narrow: it needs Cancel pressed in the window while a PowerShell session is being opened. But the failure mode is the bad kind — the app reports that it stopped and keeps working — and it is the same family as #2206, where a cancelled query returned as a successful empty result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,19 @@ That paragraph is not decoration: the release workflow copies each entry verbati
 the GitHub release body and the announcement discussion, so it is the first thing a
 prospective user reads. CI fails a pull request whose newest entry is missing it.
 
+## [1.99.6] - 2026-09-12
+
+**Pressing Cancel just as a background check starts now actually stops it.** There was a narrow window —
+the moment while the app was opening the PowerShell session it needs — where Cancel was accepted, reported
+back as cancelled, and yet the work carried on to the end anyway. Anything cancelled in that instant ran to
+completion in the background while the tab already said it had stopped. Cancel is now delivered even when
+it arrives in that window.
+
+### Fixed
+
+- Cancelling during the moment a PowerShell session is being opened now stops the script, instead of
+  reporting a cancellation while the script kept running to its natural end.
+
 ## [1.99.5] - 2026-09-12
 
 **When an update download fails, the tab now says why.** Pressing Download and having it fail showed you
--- a/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
+++ b/SysManager/SysManager.IntegrationTests/PowerShellRunnerTests.cs
@@ -583,17 +583,33 @@ public class PowerShellRunnerTests
             + $"exception {ex?.GetType().Name ?? "NONE"} ({ex?.Message ?? "-"}); {lines.Count} line(s) captured"
             + (errors.Count > 0 ? $", errors: {string.Join(" | ", errors.Take(2))}" : "");
 
-        Assert.True(sw.Elapsed < TimeSpan.FromSeconds(5),
-            "Cancellation took too long. " + diagnosis
-            + " — the exception MESSAGE is the discriminator, because both cancellation paths raise "
-            + "OperationCanceledException: \"stopped by cancellation\" means ps.Stop() interrupted the "
-            + "pipeline and only the bound failed, while \"completed on its own\" means Stop never "
-            + "interrupted anything and the script ran to its natural end. NONE means the run predates the "
-            + "post-await check entirely.");
-
         // Stopping without reporting it is its own defect: every caller's cancel branch is written around
         // OperationCanceledException, so a silent return would show a cancelled operation as a completed one.
+        // Asserted FIRST, because the two below describe what KIND of cancellation it was and neither means
+        // anything if there was none.
         Assert.True(ex is OperationCanceledException, "Cancellation did not surface as OperationCanceledException. " + diagnosis);
+
+        // The message, not the clock, and this is the assertion that actually names the defect.
+        //
+        // BlockingScript is a LOOP of 50 ms sleeps, and ps.Stop() interrupts a pipeline between statements —
+        // the blocking-call test below measured this exact shape cancelling in 2.2 s. So "completed on its
+        // own" here does NOT mean "PowerShell cannot interrupt a blocking call", which is the legitimate
+        // outcome that test pins. It means the stop was LOST: the registration fired while the instance was
+        // still NotStarted, Stop() threw InvalidOperationException, it was swallowed, and BeginInvoke then
+        // ran the script in full (#2286). The runner now re-asserts Stop after BeginInvoke to close that.
+        //
+        // This replaces a bare `elapsed < 5s`, which was the same defect measured through a proxy: it could
+        // only see a lost stop when the loss also happened to be slow, and when it fired it reported "took
+        // too long", which reads as a loaded runner. Two of the three explanations its own message offered
+        // were about timing, and the true one was not.
+        Assert.Contains("stopped by cancellation", ex!.Message, StringComparison.Ordinal);
+
+        // The bound is KEPT, below the script's own 20 seconds so a lost stop cannot pass it, and well above
+        // the ~350 ms a working stop takes. It is now a backstop rather than the discriminator: if it ever
+        // fires while the message above passes, the stop worked and the runner got slow, which is a
+        // different finding and the message says which.
+        Assert.True(sw.Elapsed < TimeSpan.FromSeconds(5),
+            "Cancellation was reported as a real stop but took too long. " + diagnosis);
     }
 
     /// <summary>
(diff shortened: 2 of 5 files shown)


laurentiu021:
**Second commit: CI found the other half, and only CI could.**

The first push closed the window where a cancel arriving during the runspace open was swallowed. The integration job then failed with something better than a re-run of the old symptom:

```
Cancellation did not surface as OperationCanceledException. elapsed 00:00:01.6876469;
token fired at 444 ms; exception PSRemotingDataStructureException (The remote pipeline has been stopped.)
```

**1.69 s against 22.5 s before the fix** — so the stop worked. Only its *reporting* was wrong. Closing the window means the stop now lands on the elevated **out-of-process** transport, where `EndInvoke` throws `PSRemotingDataStructureException` rather than the `PipelineStoppedException` an in-process runspace raises. `IsPipelineStopped` does not match that, so the catch arm never fired and a cancellation that had finally worked surfaced as a raw remoting error.

This workstation cannot reproduce it: its in-process runspace raises the type already matched. That is why the first push looked complete, and it is the clearest case this session of the non-blocking integration job earning its place — the failure it reported was not the one I was expecting and was more useful than a green run.

## Added as a separate predicate, not folded in

`IsPipelineStopped` answers "does this exception mean the pipeline was stopped", and it has negative tests to keep it honest. `PSRemotingDataStructureException` is a **general remoting protocol fault** — accepting it there would make a genuine transport breakdown read as a user cancellation.

What makes the weaker inference safe is the **call site**, not the type: the one caller is guarded by `cancellationToken.IsCancellationRequested`, so `IsRemotingTornDownByOurStop` is only ever consulted after we called `Stop()` ourselves. With no cancellation requested, the exception propagates as itself — which is what a real remoting failure must do.

Matched by type, not by that message, which is a sentence from PowerShell rather than a contract.

## Tests

Five, in the file that already owns this question:

- the fault recognised bare, and wrapped one level
- **the strict predicate still rejecting it** — this is the one that pins the separation rather than describing it
- unrelated exceptions rejected by both

## Verification after the second commit

**5580** unit tests pass, 0 failed. **29/29** `PowerShellRunnerTests`. Four projects build with 0 errors, 0 warnings. Format clean on all four. Leak scan: 43 terms over 321 diff lines, 0 hits.

Issue #1528, opened 2026-08-03
[Enhancement]: App Alerts - Move App Alerts from Monitor to Apps and align its label with its own page title ("App Installation Al...


laurentiu021:
## Problem
MainWindowViewModel.cs:175 labels the tab "App Alerts" in the Monitor group, but its own page header reads "App Installation Alerts" (AppAlertsView.xaml:23) - the nav name and the page name differ. What it does is entirely app-inventory work: AppAlertsViewModel.StartMonitoringAsync takes a baseline and watches for new installs (AppAlertsViewModel.cs:40-62), and RefreshInstalledAppsAsync lists every installed program (:96-135). Its service reads exactly the same registry trees the Uninstaller reads: AppAlertService.GetRegistryApps walks SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall and the WOW6432Node twin (AppAlertService.cs:129-142), and UninstallerService.EnrichFromRegistry walks the identical two paths (UninstallerService.cs:113-133). The Apps group (MainWindowViewModel.cs:197-200) holds App Updates, Bulk Installer, Uninstaller - install, update, remove.

## Proposed solution
Move nav-app-alerts into the Apps group and rename the nav label to "App Installation Alerts" to match AppAlertsView.xaml:23. Apps becomes 4 (App Updates, Bulk Installer, App Installation Alerts, Uninstaller) - a complete install lifecycle; Monitor loses a tab that was never about system resources.

## Rationale
"App Alerts" under Monitor reads to the aunt as "notifications from apps", which is a real and DIFFERENT tab in this app (Notification Blocker) - so the current name and placement actively mislead. Her actual need here is "did something install itself without me asking?", and she will look under Apps, beside Uninstaller, because the follow-up action to a suspicious install is to uninstall it. Putting them in the same group turns detection and remedy into one adjacency. The rename also closes a nav-vs-page-title inconsistency that only exists on this tab and Camera/Mic/Location.

## Evidence
[...]

Issue #1522, opened 2026-08-03
[Enhancement]: Notification Blocker - Move Notification Blocker from Privacy & Security to Customization


laurentiu021:
## Problem
MainWindowViewModel.cs:210 places Notification Blocker (PREVIEW) as the 8th child of Privacy & Security, the app's second-largest group. Its own page copy frames it as annoyance removal with explicitly no security dimension - NotificationBlockerView.xaml:29: "Mute the apps that nag you with pop-up notifications - update reminders, trial offers, 'rate us' prompts. Uses the same per-app switches as Windows Settings, so nothing is hacked and every change is one flip away from undone. No administrator needed." Every other tab in that group either changes a privileged security surface or removes software (Defender Tweaks, App Blocker via IFEO registry writes, File Shredder, Edge/OneDrive Remover, Debloater). Customization (MainWindowViewModel.cs:212-216) has only 3 children: Context Menu, Dark Mode Scheduler, Volume Control.

## Proposed solution
Move nav-notification-blocker into Customization, making it 4 (Context Menu, Dark Mode Scheduler, Volume Control, Notification Blocker) and returning Privacy & Security to 7. Combined with the Camera/Mic/Location move above, Privacy & Security lands back at 8 with a more coherent membership.

## Rationale
Customization is already the group for "make Windows behave the way I want it to" - how the right-click menu looks, when the theme flips, how loud each app is. "Stop this app from nagging me" is the same kind of wish and, crucially, the same risk level: no admin, per-app Windows switches, one flip to undo. Filing it under Privacy & Security both overstates the stakes (the aunt hesitates over anything in a group named Security) and hides it from where she would look. The move also fixes the two worst-balanced groups in one step instead of adding to the fuller one.

## Evidence
D:\audit-snapshot\sysmanager-audit\SysManager\SysManager\ViewModels\MainWindowViewModel.cs:210 (current placement, inDevelopment: true), :202-210 (group of 8), :212-216 (Customization, 3 children); Views\NotificationBlockerView.xaml:28-30 ("No administrator needed", annoyance framing); ViewModels\NotificationBlockerViewModel.cs:30, :80-95 (per-app AUMID search/filter, no elevation state - contrast ViewModels\AppBlockerViewModel.cs and Services\AppBlockerService.cs:27 IFEO HKLM writes, which do need admin).

## Risk / trade-off
[...]

Issue #1521, opened 2026-08-03
[Enhancement]: Sidebar - Move File Lock Detector from Monitor to Storage, and rename that group "Storage & Files"


laurentiu021:
## Problem
MainWindowViewModel.cs:176 puts File Lock Detector in Monitor. The tab is a file-path tool end to end: the user types or browses to a path, scans, and optionally ends the holding process - FileLockViewModel exposes Path, Browse (an OpenFileDialog, :60-71), ScanAsync over that path (:73-98) and KillSelected with a DialogService.Confirm (:102-131); its page copy is "Find out which process is holding a file or folder open when you get a 'file in use' error." (FileLockView.xaml:28). Meanwhile Storage (MainWindowViewModel.cs:186-188) is the smallest real group in the app with only 2 children, Disk Analyzer and Duplicate Finder - both also read-only file/folder tools ("See which folders use the most space. Read-only", DiskAnalyzerView.xaml:24; "Find files with identical content. Read-only", DuplicateFileView.xaml:24).

## Proposed solution
Move nav-file-lock into Storage (3 children: Disk Analyzer, Duplicate Finder, File Lock Detector) and rename the group label to "Storage & Files" so the group name covers per-file work, not just capacity.

## Rationale
The persona arrives at this tab from a Windows error dialog - "the file is open in another program" - while trying to delete or move a file. That is a files errand, and she will look wherever the app talks about files and folders, which is Storage. "Monitor" signals continuous watching of the machine; nothing about this tab is continuous (it has no timer and no poll loop, unlike its current Monitor siblings). The move also fixes the worst size imbalance in the nav: Storage at 2 is the only group small enough that its expander costs more than it saves.

## Evidence
D:\audit-snapshot\sysmanager-audit\SysManager\SysManager\ViewModels\MainWindowViewModel.cs:176 (Monitor placement), :186-188 (Storage group, 2 children); ViewModels\FileLockViewModel.cs:27 (Path), :60-71 (Browse), :73-98 (Scan), :102-131 (KillSelected + Confirm) - no timer, no IsActive gating (absent from MainWindowViewModel.SetActive:268-277); Views\FileLockView.xaml:27-29; Views\DiskAnalyzerView.xaml:23-25 and Views\DuplicateFileView.xaml:23-25 (sibling framing).

## Risk / trade-off
[...]

Issue #1514, opened 2026-08-03
[Enhancement]: Bandwidth Monitor - Move Bandwidth Monitor from Monitor to Network


laurentiu021:
## Problem
MainWindowViewModel.cs:178 places Bandwidth Monitor as the 7th child of the Monitor group. Its own page describes it purely in network terms - BandwidthMonitorView.xaml:27: "See how much data is moving through your network and which apps are using it. Total speed and per-app activity work with no setup. For exact per-app speeds, enable precise mode (needs administrator). Everything stays on this PC - nothing is uploaded." The Network group (MainWindowViewModel.cs:190-195) currently holds Ping, Traceroute, Speed Test, Network Repair, DNS & Hosts - every other network-facing tab in the app, and notably Speed Test, which measures the same quantity (throughput) the aunt is trying to reason about.

## Proposed solution
Move nav-bandwidth-monitor into the Network group, placed next to Speed Test (Network becomes 6 tabs: Ping, Traceroute, Speed Test, Bandwidth Monitor, Network Repair, DNS & Hosts). Monitor drops to 6.

## Rationale
The persona's trigger for this tab is "the internet is slow / something is eating my data" - she will open Network, see Speed Test, and never think to look under Monitor. Worse, Speed Test ("how fast is my connection") and Bandwidth Monitor ("what is using my connection") are the two halves of one question and are currently in different groups, so finding one does not lead her to the other. Grouping them makes the diagnosis path complete: measure the pipe, then see who is filling it.

## Evidence
D:\audit-snapshot\sysmanager-audit\SysManager\SysManager\ViewModels\MainWindowViewModel.cs:178 (current placement in Monitor), :190-195 (Network group membership); Views\BandwidthMonitorView.xaml:26-28 (network-framed page copy); ViewModels\BandwidthMonitorViewModel.cs (435 lines, per-app + total throughput, IsActive-gated poll registered in MainWindowViewModel.SetActive:275).

## Risk / trade-off
Bandwidth Monitor is one of only four VMs with a visibility-gated poll loop (MainWindowViewModel.SetActive, line 275) so it is functionally a live monitor - a maintainer could reasonably argue Monitor is the technically honest home. It is also the tab whose optional precise mode needs admin ETW, which makes it heavier than its future Network siblings. If Monitor is being thinned by several of these proposals at once, verify the residual Monitor group still reads as coherent before landing all of them.

## Affected area
[...]

PR #2311, opened 2026-09-15
feat: file three tabs under the group people look in

laurentiu021:
Three tabs were filed by **how they work internally** rather than by **what the user came to do**. No behaviour changes anywhere — only which category each sits under, plus one group rename.

Closes the mechanical part of #1514, #1521 and #1522. The fourth proposal in that set, #1528, is **deliberately not here** — see the last section.

## The three moves

| Tab | From → To | Why the old home was wrong |
|---|---|---|
| **Bandwidth Monitor** | Monitor → Network, *immediately after Speed Test* | "How fast is my connection" and "what is using my connection" are two halves of one question. Split across groups, finding one never led to the other. |
| **File Lock Detector** | Monitor → Storage, renamed **Storage & Files** | It has no timer and no poll loop — it is a one-shot scan of a path the user typed, reached from Windows' own *"the file is open in another program"* dialog. "Monitor" promises continuous watching. |
| **Notification Blocker** | Privacy & Security → Customization | Every other tab in that group changes a privileged surface or removes software. This one flips the same per-app switches Windows Settings does, needs no administrator, and is one flip from undone. |

Bandwidth Monitor is the one with a real counter-argument, and #1514 raised it: the tab genuinely *is* a live monitor — one of the few view models with a visibility-gated poll loop — so "Monitor" was defensible on mechanism. It is placed on the errand instead, because nobody whose internet feels slow thinks to look under Monitor. Its **position** is asserted rather than just its membership, since the adjacency to Speed Test is the entire point.

## The batch check #1514 asked for

#1514's risk note said: *"If Monitor is being thinned by several of these proposals at once, verify the residual Monitor group still reads as coherent before landing all of them."* Two of these three take a tab out of Monitor, so:

**Monitor after this change (5):** Process Manager · Resource History · Camera/Mic/Location · New App Alerts · Settings Watchdog.

That reads *tighter* than before, not thinner — every one of the five is genuinely something watching continuously, which is exactly what the group's own subtitle promises ("What is running, and what it is using"). The two that left are the two that were not continuous watching at all.

**Group sizes:** Monitor 7→5, Network 5→6, Storage 2→3, Privacy & Security 8→7, Customization 3→4. Storage at 2 was the app's worst imbalance — the only group small enough that its expander cost more than it saved. Total stays **58 tabs**.

One thing worth flagging for later: #1522 refers to a Camera/Mic/Location move "above" it, i.e. another proposal that also takes a tab out of Monitor. If that one lands too, Monitor drops to 4 and the coherence question should be re-asked rather than assumed — this PR does not pre-approve it.

## A decision recorded on purpose

Renaming the group to "Storage & Files" makes an obvious question askable: why isn't **File Shredder** in there too? It is also a per-file tool. It stays in Privacy & Security deliberately — shredding is a *destroy-the-traces* intent, and it is the most destructive operation in the app, so it benefits from that group's warning context. Both the code comment and the CHANGELOG say so, because otherwise it looks like an oversight.

## Testing

Three regression tests in `MainWindowViewModelTests`, following the existing `NavGroups_X_LivesInYNotZ` idiom, each **proven by reverting its own move**:

| Mutation | Result |
|---|---|
| group label back to `"Storage"` | `NavGroups_FileLock_LivesInStorageNotMonitor` **RED** |
| Bandwidth Monitor entry removed from Network | `NavGroups_BandwidthMonitor_SitsRightAfterSpeedTestInNetwork` **RED** |
| Notification Blocker removed from Customization | `NavGroups_NotificationBlocker_LivesInCustomizationNotPrivacy` **RED** |

The third needed a second attempt worth recording: Notification Blocker is the **last child** of its group, so simply deleting the line leaves the `Group(...)` call unbalanced and the mutation fails to **compile**. A build error is not a red test — it proves nothing about the assertion — so the mutation was rewritten to move the closing paren onto the previous entry, keeping the paren count identical. Anyone reverting a last-in-group entry hits the same trap.

**Suites:** unit `5691 / 5691` and the nav integration class `43 / 43`, both run twice — once before the mutations and again after the tree was restored. All four projects build 0 warnings / 0 errors; `dotnet format --verify-no-changes` clean; leak scan over all 43 patterns against 139 added lines, 0 hits (floors asserted on both counts).

**Not verified here:** that the renamed group header and the reordered rails *look* right on screen. The main workstation never launches the app, so that is a laptop check. Nothing about it is load-bearing — "Storage & Files" is 15 characters against existing group labels of 17 and 18, and no tab label changed at all, so no sidebar label budget moves.

## Docs

README's group table and ARCHITECTURE's group→view-model table both rewrote the same five rows, so the two cannot disagree with the code. SECURITY's supported line moves to 1.103.x.

## Why #1528 is not in this batch

#1528 proposes moving **App Alerts** from Monitor to Apps. It is not here because the repository already contains a **decision to the contrary**: `NavGroups_AppAlerts_LivesInMonitorNotPrivacy` asserts that tab is in Monitor, with the stated reason *"App Alerts passively watches for new installs — it belongs with the monitoring tabs."* Landing #1528 means editing that assertion, i.e. reversing a recorded call — not something to do quietly inside a batch of uncontested moves.

Two things make it more arguable than it looks, both of which I will put on the issue rather than acting on:

1. That test's other half is `DoesNotContain(..., privacy)` — it was written to settle **Monitor vs Privacy**, not Monitor vs Apps. The question #1528 raises was never the one it decided.
2. #1528 also asked to rename the label to "App Installation Alerts". That is already resolved differently: the label is now **"New App Alerts"**, which is the shorter fallback #1528's own risk note offered for the 220px rail.

So #1528 is now purely a move, blocked on one editorial call.

Refs #1514, #1521, #1522


laurentiu021:
## Update: the three tests became one, and why

The testing table above names three separate tests. They are now **one** — `NavGroups_FileTabsAreGroupedByErrand_NotByMechanism` — because the first version broke CI in a way worth writing down.

Every test in `MainWindowViewModelTests` does `var vm = new MainWindowViewModel()`, and on the test path that builds the **whole tab graph eagerly**, including the view models that launch real `winget` and `powershell` children. Adding three more constructions took the integration job from ~6 minutes to **22, twice in a row**, with nine `winget_*_hang.log` / `powershell_*_hang.log` artifacts and a host process that could no longer exit (`Exit code: -532462766`, hang dump written).

Every test passed the whole time — `652/652, failed: 0, error: 1` — and `gh pr checks` printed **pass**, because that job is `continue-on-error`. Only the duration and the log gave it away.

**Measured, not assumed:**

| Constructions | Class runtime | Exit |
|---|---|---|
| 1 | 0.9 s | clean |
| 41 | 9.7 s | forced exit |
| 43 | 27.3 s | forced exit |

Superlinear in the number of *constructions*, not assertions — so three placements against one graph costs nothing in coverage. All three mutations still red the folded test (group renamed back to "Storage"; Bandwidth removed from Network; Notification removed from Customization).

**After folding:** integration `Passed!`, 650 tests, **4m39s**, zero hang logs. Unit 5691/5691, UI 138/138, all genuinely `Passed!` read from the log rather than the check mark.

**What I did not fix here:** the leak is pre-existing — main's 40 constructions leave foreground threads too, which I proved by running the class with only the three new tests skipped (still rc=1). This PR just stops adding to it. Filed as #2312, along with a second thing that surfaced while comparing baselines: `PowerShellRunnerTests.RunAsync_CancellingASingleBlockingCall_CannotInterruptItButStillReportsIt` **failed on main** in run `34941054173` (`failed: 1, succeeded: 648`) and was reported as a passing check by the same masking.

One extra commit here fixes stale comments my own change introduced: `NavItems_ContainAll58` groups its 58 assertions under comments naming each group and its size, and those still described the old layout. The assertions passed regardless because the list is flat, so the comments were quietly wrong — re-derived to 58 ids, no duplicates.

Issue #2381, opened 2026-09-22
[Bug]: Disk Analyzer, Duplicate Finder and Large Files keep their own tree walk, and it misses the root guard


laurentiu021:
## Problem

Unifying the cleanup walkers surfaced three more: `DiskAnalyzerService`, `DuplicateFileService` and
`LargeFileScanner` each keep their own `Stack<string>` tree walk. They are exempted from
`OnlySafeFileWalk_WalksATreeItMightDeleteFrom` by a named list, deliberately and visibly, rather than left to
be found again — this issue is that list's other half.

They were left out because they need something `SafeFileWalk` does not offer: a **per-directory hook**. Each
one reports throttled progress naming the folder currently being scanned, at the end of every directory:

```csharp
// LargeFileScanner
var now = Environment.TickCount64;
if (now - lastReport >= 200)
{
    progress?.Report(new LargeFileProgress(scanned, bytesScanned, cur));
    lastReport = now;
}
```

A flat sequence of file paths cannot carry that. Migrating them means giving the shared walk a
"finished a directory" callback, which is a design decision worth taking on its own rather than inside a
refactor of the delete paths.

## Why it matters that they are walks and not just scans

Two of the three feed a delete. Large Files and Duplicate Finder both present what they found with a way to
remove it, so the same reparse-point reasoning applies as for the cleaners — and their handling differs from
the rule now shared by everything else:

- **The traversal ROOT is not guarded.** `LargeFileScanner` pushes `rootPath` and only tests reparse points
  on child directories. The user picks the folder to scan, so a junction there walks straight into its target
  and offers files outside the selected folder for deletion. Every cleanup walker had this same gap and it
  was closed in all of them; here it is still open.
- **Link FILES are not skipped.** `FileInfo.Length` for a link reports its target, so a link can be listed
  as one of the biggest files on the drive, or as a duplicate of the thing it points at. Deleting it takes
  the link and frees nothing — the number shown was never real.
- **The `MoveNext` half is *not* an exposure here**, unlike `TuneUpService`'s copy. All three call
  `Directory.GetFiles` / `GetDirectories`, which return arrays, so the throw lands on the call — and all
  three guard the call. Stated rather than left implied, so the next person does not assume the worst case
  and does not assume the best one either.

## Proposed solution

Add a per-directory callback to `SafeWalkOptions`, then route all three through it:

```csharp
internal Action<string>? DirectoryFinished { get; init; }
```

[...]

PR #2382, opened 2026-09-22
fix: one directory walk for every service that deletes what it finds

### LINKED ISSUE #2376: [Bug]: Deep Cleanup — its file walk and delete drifted from the shredder on two safety details
## Problem

Deep Cleanup and File Shredder both walk a tree and destroy what they find, and their walkers have drifted
apart on two details the shredder gets right and documents.

**1. The walk does not skip link files, though the single-file branch of the same helper does.**

`EnumerateTargets` guards a path entry that names a file:

```csharp
// Guarded like a traversal root: a symlink here would make the caller delete its target,
// outside the bucket's tree. IsReparsePoint fails safe on an access error.
if (!IsReparsePoint(path) && Matches(path, patterns)) yield return path;
```

The directory walk in the same helper yields every file `Directory.EnumerateFiles` returns, with no such
check. `EnumerateFiles` skips reparse-point *directories* and the traversal root, never a reparse-point
*file*. The shredder's walker does check, deliberately:

```csharp
// Skip symlink/hardlink files: shredding would overwrite the LINK TARGET
// (which may live outside the selected folder), so only real files in the
// tree are collected — mirrors the reparse-point skip on directories below.
```

Either the guard on the single-file branch is unnecessary or the walk is missing it. They cannot both be
right, and the inconsistency is what makes this worth closing: the next person to read this helper has to
guess which half states the rule.

`File.Delete` on a symlink removes the link rather than its target, so this is not data loss — the shredder
notes the same thing. What is *not* established is the reported figure: `SafeLength` uses `FileInfo.Length`,
and whether that returns the link's own size or its target's needs checking against .NET 10 rather than
assumed. If it follows the link, a bucket containing a link to a large file counts those bytes into the scan
total and into `BytesFreed` for a delete that reclaimed nothing. Worth measuring as part of the fix — file
symlink creation needs developer mode or elevation, so it did not get measured while filing this.

**2. The delete wipes every attribute rather than clearing the one that blocks a delete.**

```csharp
var len = SafeLength(file);
File.SetAttributes(file, FileAttributes.Normal);
File.Delete(file);
```

`FileAttributes.Normal` is not a mask — it replaces the set, dropping Hidden, System and Archive along with
the ReadOnly bit that is the only one standing in the delete's way. When the delete then fails, and in these
buckets it often does, the file stays on disk having silently lost its attributes. The Explorer cache bucket
is the clearest case: `thumbcache_*.db` and `iconcache_*.db` are hidden system files that Explorer holds
open, so every clean strips their attributes and then cannot remove them.

The shredder, again, does the narrow thing:

```csharp
var attrs = File.GetAttributes(filePath);
if (attrs.HasFlag(FileAttributes.ReadOnly))
    File.SetAttributes(filePath, attrs & ~FileAttributes.ReadOnly);
```

## Steps to reproduce

For (2): run Deep Cleanup with "Explorer thumbnail & icon cache" selected while Explorer is running (which
it always is). Afterwards `%LOCALAPPDATA%\Microsoft\Windows\Explorer` holds the same cache files with
Hidden and System cleared, and the run reports them as errors rather than deletions.

For (1): place a symbolic link inside any bucket path and compare what the scan counts with what the clean
actually reclaims.

## Expected behavior

- The file walk skips reparse-point files, exactly as the shredder's does and as `EnumerateTargets`
  already does for a path entry that names one. Sharing one rule is the point: two branches of the same
  helper should not disagree about what a link is.
- Clearing an attribute to allow a delete clears `ReadOnly` and leaves the rest, so a delete that fails
  leaves the file as it found it.
- A test pins whichever way `FileInfo.Length` treats a link, so the reported figure is a measured fact
  rather than an assumption either way.

## Affected tab

Deep Cleanup

### LINKED ISSUE #2380: [Bug]: One-Click Tune-Up — one unreadable temp entry silently ends the whole temp clean
## Problem

There are five hand-copied tree walkers in the services, and the fix for a known enumerator defect landed in
one of them. `TuneUpService` has the other lazy walker, without the fix.

`DeepCleanupService.EnumerateFiles` wraps the iteration because the enumerator can throw from `MoveNext`
rather than from the call that created it, and the comment records that this was measured rather than
reasoned about:

```csharp
// Wrap iteration because the enumerator can throw from MoveNext() rather than from the call
// above — a directory that stops being one mid-walk, or an entry that becomes unreadable.
using var enumerator = files.GetEnumerator();
while (true)
{
    string? item;
    try { if (!enumerator.MoveNext()) break; item = enumerator.Current; }
    catch (IOException) { break; }
    catch (UnauthorizedAccessException) { break; }
    yield return item;
}
```

`TuneUpService.EnumerateFilesSkippingReparsePoints` guards only the call, then iterates bare:

```csharp
try { files = Directory.EnumerateFiles(cur); } catch (IOException) { continue; } catch (UnauthorizedAccessException) { continue; }
...
foreach (var file in files)
    yield return file;
```

`EnumerateDirectoriesSkippingReparsePoints`, immediately below it, has the same shape.

A throw from `MoveNext` therefore escapes the walker, escapes the delete loop in `CleanTempFiles`, and is
caught by the per-directory handler wrapping the whole block:

```csharp
catch (IOException ex)
{
    Log.Warning("TuneUp temp cleanup error in {Dir}: {Error}", dir, ex.Message);
}
```

So one unreadable entry part-way through `%TEMP%` ends the cleanup of that entire directory. Not a crash and
not an infinite loop — `foreach` disposes the enumerator, so this is not the spin the Deep Cleanup comment
describes — but the Tune-Up then reports the bytes it had freed up to that point as its result, with nothing
saying it stopped early. The user is told the temp clean finished. The only record that it did not is a
Warning in the log file.

`%TEMP%` is the single most churn-heavy directory on a Windows machine, and it is being walked while other
processes create and delete files in it, which is exactly the condition that makes an entry vanish between
being listed and being read.

## Which walkers are exposed

| Walker | Enumeration | Throws from `MoveNext` |
|---|---|---|
| `DeepCleanupService.EnumerateFiles` | `Directory.EnumerateFiles` (lazy) | guarded, skips the directory |
| `TuneUpService.EnumerateFilesSkippingReparsePoints` | `Directory.EnumerateFiles` (lazy) | **unguarded** |
| `TuneUpService.EnumerateDirectoriesSkippingReparsePoints` | `Directory.EnumerateDirectories` (lazy) | **unguarded** |
| `BrowserCleanerService.SafeEnumerateFiles` | `Directory.GetFiles` (eager array) | not possible |
| `FileShredderService.EnumerateFilesSafe` | `DirectoryInfo.GetFiles` (eager array) | not possible |
| `ShortcutCleanerService.EnumerateLnkFilesSafe` | `Directory.GetFiles` (eager array) | not possible |

## Steps to reproduce

Walk `%TEMP%` while another process is deleting files in a subdirectory of it — or point the walker at a
directory that is replaced by a file mid-walk, which is the case the Deep Cleanup comment says it measured
across ten attempts. The exception surfaces as a single `Warning` and the remaining files in that temp root
are never visited.

## Expected behavior

- Both `TuneUpService` walkers guard `MoveNext` the way `DeepCleanupService` does, ending that directory
  rather than the whole root.
- `CleanTempFiles` counts a walk that ended early so the result can say so, instead of reporting a partial
  figure as a completed clean. The Tune-Up card already distinguishes "skipped" from "done" for the Recycle
  Bin; the temp step has no equivalent.

## Related

The same five walkers are the subject of #2376, which found the reparse-point-file rule present in one of
them. That is the second rule to have been copied rather than shared. There is a real refactor underneath
both issues — one safe-walk helper with the reparse boundary, the exclusion list, the cancellation check and
the `MoveNext` guard in a single place — and each copied walker is a place for the next rule to go missing
too. Worth deciding whether these two bugs are fixed in place or fixed by removing the duplication.

## Affected tab

Dashboard (One-Click Tune-Up, and the Quick Cleanup shortcut that shares `CleanTempFilesAsync`)

laurentiu021:
Five hand-copied tree walkers, four private copies of `IsReparsePoint`, and two safety rules that only ever reached one copy. Each copy carried a comment saying it mirrored the others — which is exactly how a rule added to one goes missing from the rest, and no per-service test could see it, because every copy passed its own.

Closes #2376
Closes #2380

## The two rules that had drifted

| Rule | Had it | Missing from |
|---|---|---|
| Never yield a reparse-point **file** (every copy already refused to descend into a reparse-point **directory**) | `FileShredderService` | `DeepCleanupService`, `TuneUpService`, `BrowserCleanerService`, `ShortcutCleanerService` |
| Absorb a listing that throws from `MoveNext`, not just from the call that created the enumerator | `DeepCleanupService` | `TuneUpService` (the one other lazy walker) |

**The file rule (#2376).** `File.Delete` on a symlink takes the link, so no data was lost. But `FileInfo.Length` reports the *target's* size, so a cleanup bucket billed the user for bytes no delete could reclaim. The two halves of `EnumerateTargets` also disagreed with each other: its single-file branch guarded against a link, its walk did not.

**The enumerator rule (#2380).** `TuneUpService` guarded the call and iterated bare, so one unreadable entry escaped the walk and ended the cleanup of an entire temp root — while `CleanTempFiles` reported the bytes freed up to that point as a finished clean. `%TEMP%` is the busiest directory on the machine and other processes write it during the sweep, which is exactly when an entry stops being readable between being listed and being opened.

## The shared walk

`Helpers/SafeFileWalk` owns four rules: never enter a reparse point (root, directory or file), skip the excluded subtrees, honour the token between directories, and enclose the iteration.

`EnumerateGuarded` returns a `List` rather than yielding, and that is load-bearing rather than incidental — **a `yield return` cannot sit inside a `try` that has a `catch`**, which is precisely why four of the five copies never enclosed their own iteration. `Files()` stays lazy and holds one directory at a time, so a walk over `%TEMP%` never materialises it.

`SafeWalkOptions` is a record, not optional parameters: with a `params string?[]` last, a caller passing an exclusion path positionally would have bound it to the search pattern silently.

Migrated: `FileShredderService`, `DeepCleanupService`, `TuneUpService`, `BrowserCleanerService`, `ShortcutCleanerService`, `CleanupPreScanService`. **Net −340 lines**, and one `File.GetAttributes` syscall per entry is gone — the copies asked Windows for each subdirectory's attributes a second time; the shared walk reads them off the `FileInfo`/`DirectoryInfo` the enumeration already populated.

## Also fixed here (#2376's second half)

```diff
-File.SetAttributes(file, FileAttributes.Normal);
+var attrs = File.GetAttributes(file);
+if ((attrs & FileAttributes.ReadOnly) != 0)
+    File.SetAttributes(file, attrs & ~FileAttributes.ReadOnly);
```

`Normal` is not a mask — it replaces the set — so it dropped Hidden, System and Archive along with the ReadOnly bit that is the only one blocking a delete, and a file whose delete then failed stayed on disk without them. The Explorer cache bucket does this on **every** run: `thumbcache_*.db` and `iconcache_*.db` are hidden system files Explorer holds open. Now clears ReadOnly only, which is what the shredder already did.

## Three scanners are NOT migrated, and say so

`DiskAnalyzerService`, `DuplicateFileService` and `LargeFileScanner` report throttled progress per **directory**, which this walk cannot carry yet. Giving it that hook is #2381 — which also records that `LargeFileScanner` does not guard its traversal root.

They are a **named list in the guard, asserted to stay live**: a name whose file no longer walks a tree fails the test, so the exemption cannot rot into a free pass. The guard found all three on its first run, which is how they came to be filed at all.

## Guards

`OnlySafeFileWalk_WalksATreeItMightDeleteFrom` bans a private `IsReparsePoint` and a `Stack<string>`/`Stack<DirectoryInfo>` walk over a file listing, across every source file but `SafeFileWalk` and the three exemptions. Proven red both ways: it found the three scanners unprompted, and re-adding a private `IsReparsePoint` to `ShortcutCleanerService` fails it by name.

**Three existing guards had to be re-derived rather than deleted**, and each failed with its own vacuity message telling me exactly that:

- `EveryTempTreeWalkerCall_PassesBothExtractionExclusions` read six repeated call-site arguments. The exclusions now live in one named `SafeWalkOptions` per service, so it checks those two declarations **and** that no call passes a bare `new SafeWalkOptions()` instead. Floor unchanged at six calls, because none disappeared.
- `EveryFilteredCleanupBucket_IsFilteredInBothScanAndClean` counted 2 uses of the private walk (declaration + call); it now requires exactly 1 `SafeFileWalk.Files` call and asserts it sits inside `EnumerateTargets`, where the pattern filter is applied.
- `TheCleanupWalk_AbandonsADirectoryWhoseEnumeratorThrows` pinned a `try`/`catch` around `MoveNext`. **A mutation proved that shape redundant in the new helper** — removing it changed no behaviour, because the enclosing `try` already absorbed the throw. So the redundant code is gone and the guard now asserts what does the work: that the `foreach` over `enumerate()` is *inside* the try. Asserting only that `enumerate()` appears inside a try would have passed on the defect itself, since `try { source = enumerate(); } catch {…} foreach (… in source)` satisfies it while leaving every `MoveNext` unguarded — the exact shape `TuneUpService` had.

## Tests

15 in `SafeFileWalkTests`, which is where the 8 cases previously split between `TuneUpServiceTests` and `ShortcutCleanerViewModelTests` now live. The walk is one thing, so it is tested in one place — testing it per service is what let two copies diverge.

Red before / green after:

- `CleanAsync_WhenADeleteFails_LeavesTheFilesAttributesAlone` — restoring `FileAttributes.Normal` fails it with *"Hidden was stripped from a file that survived"*. Deterministic via an open handle: no privileges, no timing.
- `EnumerateGuarded_KeepsWhatItReadBeforeMoveNextThrew` and `_DoesNotRetryAnEnumeratorThatThrew` — guarding the call instead of the iteration makes both throw `IOException`, and turns the architecture guard red at the same time.
- The reparse-point-**file** case needs a file symlink, which needs Developer Mode or elevation. It self-skips on this machine and runs in CI. The privilege-free half of that rule — `IsReparsePoint` failing closed on an unreadable path — is asserted unconditionally, so the boundary is never wholly untested.

## Verification

- 5911 unit tests, 0 failed
- All four projects build, 0 warnings
- ARCHITECTURE.md documents the new helper, including why the exemption list exists


laurentiu021:
CI evidence for the one rule that cannot be proven on this machine.

The three reparse-point cases now report a skip rather than returning silently, so the run summary answers whether they executed. **`skipped: 0` in CI** against `skipped: 3` locally — so `Files_DoesNotYieldAReparsePointFile_AndReportsIt` ran on a real symlink on the runner and passed, which is the only place the #2376 file rule can be exercised.

Worth being precise about what that is and is not: it is positive coverage of a new rule on real symlinks, not a red-then-green regression pin. The old code had no equivalent API to fail against — its walk simply yielded the link — so there is no "before" for this test to be red against. The other three fixes in this PR do have that proof, and it is listed in the description.

